PR #5, opened 2026-03-03
Updating readme with latest changes

gavin-aguiar:
Updating readme


ahmedmuhsin (commented):
I believe deployment to azure is currently possible if a custom image is used. Should we detail that here? If we don't want customers to deploy yet we should explicitly say that deployment to Azure is not supported.

PR #19, opened 2026-04-01
Implement triggers-only programming model

ahmedmuhsin:
## Summary

Adopt a triggers-only programming model, removing framework-injected input/output bindings in favor of typed trigger handlers. The SDK is organized into two tiers: **Core Triggers** (data passthrough, zero external deps) and **Extension Triggers** (SDK client injection, isolated deps via blank import).

### Programming Model
- **Typed handler signatures**  each trigger type has a compile-time-checked handler type (e.g. `TimerHandler`, `ServiceBusHandler`, `EventHubHandler`). No more `interface{}` and runtime reflection errors.
- **Remove input/output binding infrastructure**  extensions directory, converter registry, blank-import pattern, reflection-based parameter matching, and all output binding methods on builders.
- **Add `context.Context` + `error` return** to all non-HTTP handler signatures.
- **`name` parameter** now sets the function display name (was previously ignored).
- **Consolidate `RegisterFunction`**  removed the parameterless `RegisterFunction(f, b)` variant; single exported method `RegisterFunction(name, f, b)` always requires an explicit name.

### Two-Tier Trigger Architecture

**Core Triggers** (`sdk/`): HTTP, Timer, CosmosDB, ServiceBus, EventHub, EventGrid
- Data passthrough  host serializes trigger payload into the gRPC `InvocationRequest`, worker deserializes into typed Go structs
- Zero external Azure SDK dependencies  only `encoding/json` needed
- Typed handler aliases provide compile-time safety

**Extension Triggers** (`triggers/`): Blob (and future Queue, Table, etc.)
- SDK client injection  handler receives an authenticated Azure SDK client (e.g., `*blob.Client`) scoped to the triggering resource
- Heavy deps (`azblob`, `azidentity`) isolated in `triggers/<name>/`  users who don't need blob get zero overhead
- `ClientFactory` registry pattern  extensions register via `init()`, activated by blank import

This split follows the same rationale as the .NET Isolated worker extensions model but avoids over-abstracting core triggers that don't need isolated dependencies. Decision criteria: if the payload is bounded and already in the gRPC message, it's a Core Trigger. If it's potentially unbounded or requires a live SDK client, it's an Extension Trigger.

### Blob Trigger
- **`*blob.Client` from Azure SDK**  avoids shuttling blob content over gRPC for large blobs.
- **`ClientFactory` registry**  `triggers/blob` registers its factory via `init()` (same pattern as `database/sql` drivers). Users import with `_ "github.com/.../triggers/blob"`.
- **Auto-detect auth**  connection string, `UseDevelopmentStorage=true`, managed identity (system-assigned + user-assigned via `__clientId`), and `__accountName`/`__blobServiceUri`/`__serviceUri` suffixed env vars. Aligned with the Java worker's `BaseBlobHydrator` pattern.
- **Service client caching** per connection setting with double-checked locking.
- **Handler signature validation** at registration time  `app.Blob()` validates that the handler is `func(context.Context, T) error` and panics immediately with a descriptive message for wrong signatures.

### Testing
- **Integration test framework** using pytest + uv with per-test env var isolation.
- **Docker Compose emulators**  Azurite, Service Bus, Event Hub, Cosmos DB.
- **9 integration tests**  all trigger types verified end-to-end against emulators.
- **Poll-and-retry** approach for Event Hub trigger (handles slow partition claims).

### Documentation
- **README.md**  new "Trigger Model" section explaining Core vs Extension triggers with decision criteria table and code examples.
- **TECHNICAL_SPEC.md**  Section 4 rewritten with correct terminology (`ClientFactory` not `RegisterConverter`), correct paths (`triggers/blob` not `sdk/extensions/blob`), correct handler signatures, and examples for both tiers.
- **Code comments**  banner comments in `sdk/app.go` demarcating the two tiers, doc blocks in `sdk/handlers.go` and `sdk/factories.go` explaining design rationale.
- **READINESS.md**  checklist of before-Preview and before-GA items for tracking remaining work.

### Code Quality
- Consistent naming aligned with Azure Functions ecosystem (`HttpTrigger`, `CosmosDBTrigger`, `BlobTrigger`, etc.)
- Removed dead code (unused types, functions, deprecated `ioutil`)
- Fixed `HashFunctionID` collision risk (includes name + trigger type)

## Why

- All current Go users are on custom handlers and already create their own service clients. This model preserves that pattern while eliminating boilerplate (manual `function.json`, HTTP envelope parsing, untyped trigger data).
- The two-tier architecture (Core vs Extension) keeps the SDK lean  core triggers have zero external deps and typed signatures, while extension triggers isolate heavy Azure SDK dependencies for opt-in use only.
- Reduces SDK surface area by ~70%.

## Handler Signatures

| Trigger | Handler Type |
|---|---|
| HTTP | `http.HandlerFunc` |
| Timer | `func(context.Context, TimerInfo) error` |
| Service Bus | `func(context.Context, ServiceBusMessage) error` |
| Event Hub | `func(context.Context, EventHubMessage) error` |
| Cosmos DB | `func(context.Context, []CosmosDocument) error` |
| Event Grid | `func(context.Context, EventGridEvent) error` |
| Blob | `func(context.Context, *blob.Client) error` (import `triggers/blob`) |

## Migration from Custom Handlers

| | Custom Handler (today) | Native Triggers-Only (this PR) |
|---|---|---|
| Files per function | `function.json` (manual JSON) | None  defined in code |
| Trigger data | `json.RawMessage`  parse yourself | Typed struct |
| HTTP plumbing | Manual `ListenAndServe`, port env var | `worker.Start(app)` |
| Service clients | Create your own | **Create your own (no change)** |
| Compile-time safety | None | Handler type checked by compiler |


sdk/bindings/http.go
```diff
@@ -1,9 +1,10 @@
 package bindings
 
-const HttpBindingType BindingType = "httpTrigger"
+// HttpTriggerBindingType is the binding type constant for HTTP triggers.
+const HttpTriggerBindingType BindingType = "httpTrigger"
```
harshivcodes: nit; naming: `HttpTrigger `instead of `HttpTriggerBindingType` since intellisense will show that its of type `BindingType`

sdk/bindings/http.go
```diff
@@ -17,14 +18,14 @@ type HttpTrigger struct {
...
-func (c *HttpTrigger) GetBindingType() BindingType { return HttpBindingType }
+func (c *HttpTrigger) GetBindingType() BindingType { return HttpTriggerBindingType }
 
 func (c *HttpTrigger) ToBinding() Binding {
 	return Binding{
 		Name:      c.Name,
 		Type:      string(c.GetBindingType()),
 		Direction: "in",
-		HTTPBinding: &HTTPBinding{
+		HttpBinding: &HttpBinding{
```
harshivcodes: for separate pr, I was looking up golang naming conventions it might be worthwhile to enforce a linting check in our pipeline. We'll have to revisit: `Http`, `CosmosDb` etc.

examples:
```
type APIClient struct{}
type HTTPServer struct{}
type URLParser struct{}
type UserID string
```

sdk/bindings/blob.go
```diff
@@ -1,45 +1,23 @@
 package bindings
 
-const BlobBindingType BindingType = "blobTrigger"
-const BlobInputType BindingType = "blob"
-
+// BlobTriggerBindingType is the binding type constant for Blob triggers.
+const BlobTriggerBindingType BindingType = "blobTrigger"
```
gavin-aguiar: Can you create a unit test for this file?

sdk/bindings/eventgrid.go
```diff
@@ -3,13 +3,8 @@ package bindings
...
 
 const EventGridTriggerBindingType BindingType = "eventGridTrigger"
-const EventGridOutputType BindingType = "eventGrid"
-
-// EventGridBinding is the JSON representation for EventGrid bindings.
-type EventGridBinding struct {
-	TopicEndpointUri string `json:"topicEndpointUri,omitempty"`
-	TopicKeySetting  string `json:"topicKeySetting,omitempty"`
-}
+// EventGridBinding is the JSON representation for EventGrid trigger bindings.
```
gavin-aguiar: NIT: Missing like between constants and type

sdk/bindings/http.go
ahmedmuhsin: Agreed, this is worth doing. golangci-lint with stylecheck (ST1003) would catch the naming issues. We've adopted Go conventions (HTTP*, CosmosDB*) in this PR so the linter would pass. Will add to the CI pipeline as a follow-up.

sdk/bindings/blob.go
ahmedmuhsin: Done.

sdk/bindings/common.go
```diff
@@ -44,8 +37,8 @@ func (b Binding) MarshalJSON() ([]byte, error) {
 	var sub interface{}
```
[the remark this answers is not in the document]
ahmedmuhsin: Done.

sdk/bindings/http.go
ahmedmuhsin: Done.

sdk/app.go
```diff
@@ -660,63 +370,138 @@ func (b *ServiceBusTopicFunctionBuilder) Cardinality(cardinality string) *Servic
...
+	trigger := &bindings.BlobTrigger{
+		Name: "blob",
+	}
+
+	rf := app.registerFunction(name, f, trigger)
+
+	// Look up the globally registered factory for blob triggers
+	if factory, ok := GetClientFactory(string(bindings.BlobTriggerBindingType)); ok {
+		rf.ClientFactory = factory
 	}
```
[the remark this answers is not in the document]
ahmedmuhsin: Done.

harshivcodes (approved):

tests/integration/conftest.py
```diff
@@ -0,0 +1,138 @@
+"""Shared fixtures for integration tests."""
```
gavin-aguiar: Why tests are in python?

ahmedmuhsin: What's the reasoning for this? Are you saying the tests should be implemented in Go or have no tests at all?

tests/integration/test_blob_trigger.py
```diff
@@ -0,0 +1,45 @@
...
+    "TableEndpoint=http://127.0.0.1:10002/devstoreaccount1;"
+)
+
+ENV = {
+    "AzureWebJobsStorage": AZURITE_CONN_STR,
+    "FUNCTIONS_WORKER_RUNTIME": "golang",
+}
+
+
+def test_blob_trigger_fires(func_host):
```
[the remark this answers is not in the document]
ahmedmuhsin: It's a pytest fixture defined in `conftest.py`. It's a factory function that builds the Go sample, starts `func.exe` with the given env vars, waits for worker initialization, and returns a process handle with log assertion helpers. Cleanup (kill process) happens automatically via the fixture's yield/teardown.

PR #26, opened 2026-04-10
Flex consumption proxy and specialization tests

ahmedmuhsin:
## Summary

Implements and tests the gRPC proxy for Azure Functions Go worker flex consumption support. The proxy handles placeholder mode during cold start, then spawns the user's compiled Go binary on specialization.

## Proxy (proxy/main.go)

A ~380-line gRPC proxy that sits between the Azure Functions host and the user's Go app during flex consumption placeholder mode.

### Lifecycle
1. **Exec bypass**: If the app binary exists at startup, the proxy replaces itself via `syscall.Exec` — zero overhead, host talks directly to the app
2. **Placeholder mode**: Proxy connects to the host, handles `WorkerInitRequest`, heartbeats, and metadata requests (returns empty)
3. **Specialization**: On `FunctionEnvironmentReloadRequest`, proxy applies env vars to its own process, spawns the child app, captures its capabilities, and sends `FunctionEnvironmentReloadResponse` with the child's real capabilities
4. **Steady state**: Transparent gRPC bridge — all messages forwarded between host and child
5. **Child death**: Proxy exits with the child's exit code; host restarts; exec bypass kicks in

### Key design decisions
- Child capabilities forwarded via `FunctionEnvironmentReloadResponse` (`replace` strategy) — proxy doesn't dictate capabilities
- FERR env vars applied to proxy process via `os.Setenv` — keeps proxy and child aligned
- `HandlesWorkerTerminateMessage` capability enables graceful shutdown
- `FUNCTION_APP_DIRECTORY` and `FUNCTION_APP_NAME` env vars for configurability
- No dedicated mode — proxy is flex consumption only

### Functional tests
- `TestFlexConsumptionPlaceholderPing` — proxy handles placeholder mode, host returns 200
- `TestFlexConsumptionHttpTriggerProxy` — full lifecycle: placeholder → deploy → specialize → HTTP trigger

### Benchmarks
- `BenchmarkHttpTriggerDirect` — baseline with exec bypass (app prebaked in image)
- `BenchmarkHttpTriggerProxy` — production path through gRPC proxy

### Benchmark results (AMD EPYC 7763, Docker Desktop, 10s benchtime)

| Mode | Latency | Delta |
|------|---------|-------|
| Direct (exec bypass) | ~5.4 ms/op | baseline |
| Proxy (gRPC bridge) | ~5.7 ms/op | +0.3 ms (+5%) |

### Unit tests (proxy/main_test.go)
- `TestAppBinaryPath_Default` / `TestAppBinaryPath_CustomName`
- `TestAppBinaryPath_PicksUpSetenv` — verifies FERR env propagation
- `TestSetenvOverridesExisting` — verifies no duplicate env entries

## Files changed
- `proxy/main.go` — Proxy implementation
- `proxy/main_test.go` — Unit tests
- `proxy/README.md` — Documentation
- `tests/consumption/` — E2E tests, benchmarks, Dockerfiles, README
- `worker.config.json` — Removed Arm64, X86 from supported architectures


proxy/main.go
```diff
@@ -0,0 +1,420 @@
...
+
+func main() {
+	log.Println("Go Worker Proxy Starting...")
+
+	// If the real app exists, exec into it directly. This handles the post-
+	// specialization case: after deployment places the real binary at
+	// /home/site/wwwroot/app, the host restarts and starts the "proxy" again
+	// (since defaultExecutablePath points here). Instead of running as a proxy,
+	// we replace ourselves with the real app. Zero overhead.
+	appDir := os.Getenv("FUNCTION_APP_DIRECTORY")
```
gavin-aguiar: I don't think we should add this. This will open a whole new set of errors like directory not found, permissions issue, logging etc. Lets also default it to /home/site/wwwroot

proxy/main.go
```diff
@@ -0,0 +1,420 @@
...
+						"RawHttpBodyBytes":                  "true",
+						"RpcHttpTriggerMetadataRemoved":     "true",
+						"UseNullableValueDictionaryForHttp": "true",
+						"HandlesWorkerTerminateMessage":     "true",
+					},
+					WorkerVersion: "1.0.0",
+				},
+			},
+		}
+		p.hostStream.Send(response)
```
gavin-aguiar: The main loop and the child -> host bridge goroutine both call hostStream.Send() without synchronization. gRPC streams are not safe for concurrent sends .  Maybe we can add a send mutex or funnel all host sends through a single goroutine

proxy/main.go
```diff
@@ -0,0 +1,420 @@
...
+
+func (p *Proxy) handleHostMessage(msg *pb.StreamingMessage) {
+	// Check if we are specialized/Proxying
+	p.mutex.Lock()
+	childReady := p.childStream != nil
+	specializing := p.isSpecializing
+	p.mutex.Unlock()
+
+	if childReady {
+		if err := p.childStream.Send(msg); err != nil {
```
gavin-aguiar, line 208: After childConnected closes, both specialize() and handleHostMessage() can send to childStream simultaneously. Also, host messages may arrive before the replayed init completes, violating protocol ordering.
Don't forward host traffic until childInitDone.

proxy/main.go
```diff
@@ -0,0 +1,420 @@
...
+			log.Printf("Error sending to child: %v", err)
+		}
+		return
+	}
+
+	// If we're specializing (env reload in progress),
+	// wait for the child to connect before forwarding.
+	if specializing {
+		log.Printf("Specializing - waiting for child before forwarding: %T", msg.Content)
+		<-p.childConnected
```
gavin-aguiar: Can we add a timeout here in the case the child never connects. Same in the specialize method.

proxy/main.go
```diff
@@ -0,0 +1,420 @@
...
+}
+
+// Implement FunctionRpcServer.EventStream (Accepts connection from Child)
+func (p *Proxy) EventStream(stream pb.FunctionRpc_EventStreamServer) error {
+	log.Println("Child connected to Proxy")
+	p.mutex.Lock()
+	p.childStream = stream
+	p.mutex.Unlock()
+
+	close(p.childConnected)
```
gavin-aguiar: This unconditionally closes the channel. A second child connection panics. Also, if the child disconnects, childStream is never cleared, so messages route to a dead stream.

proxy/main.go
```diff
@@ -0,0 +1,420 @@
...
+	}
+	p.listener = lis
+	log.Printf("Proxy listening on %s", lis.Addr().String())
+
+	p.server = grpc.NewServer()
+	pb.RegisterFunctionRpcServer(p.server, p)
+
+	go func() {
+		if err := p.server.Serve(lis); err != nil {
+			log.Fatalf("Local server failed: %v", err)
```
gavin-aguiar: This will treat a clean shutdown as a crash. Change the log to a print "Server Stopped"

proxy/main.go
ahmedmuhsin: Removed. The exec bypass hardcodes /home/site/wwwroot and the specialization path uses FunctionAppDirectory from the host's FERR. Also renamed FUNCTION_APP_NAME to FUNCTIONS_APP_BINARY_NAME to avoid confusion with the Azure Functions app name.

proxy/main.go
ahmedmuhsin: Protocol ordering prevents actual overlap in practice (the run() goroutine only writes to hostStream during placeholder, and after specialization it only writes to childStream while the bridge goroutine writes to hostStream). But that reasoning is fragile and easy to break with future changes. Added a sendToHost helper method that wraps hostStream.Send with a mutex. All host sends now go through this single method.

proxy/main.go
ahmedmuhsin: Changed the specializing path in handleHostMessage to wait on childInitDone instead of childConnected. This ensures the child has received and responded to the replayed WorkerInitRequest before any host messages are forwarded to it, preserving protocol ordering. Also removes the concurrent childStream.Send concern since specialize() finishes writing to childStream before childInitDone closes.

proxy/main.go
ahmedmuhsin: The host already enforces initializationTimeout and environmentReloadTimeout from worker.config.json. If the child doesn't connect or respond in time, the host kills the proxy process. Adding a separate timeout at the proxy level would be redundant and could conflict if someone configures a longer host timeout. The cmd.Wait() goroutine also covers the crash case, exiting the proxy immediately if the child dies.

proxy/main.go
ahmedmuhsin: Used sync.Once to protect the channel close. A second child connection won't panic. The dead stream issue is handled by child lifecycle management: cmd.Wait() goroutine calls os.Exit when the child dies, so the proxy doesn't continue sending to a dead stream.

proxy/main.go
ahmedmuhsin: Changed to log.Printf. The server returns nil on graceful stop (via defer p.server.Stop()) so this only fires on actual errors now.

proxy/main.go
```diff
@@ -0,0 +1,426 @@
...
+	cmd.Dir = dir
+	if cmd.Dir == "" {
+		cmd.Dir, _ = os.Getwd()
+	}
+
+	// Forward stdout/stderr for debugging
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+
+	log.Printf("Starting Child Worker: %s %v in %s", workerPath, args, cmd.Dir)
```
harshivcodes, line 345: I see quite a few log.print calls, these might add overhead on host - child messages, it might be best to consolidate log messages and keep the ones that are useful for debugging.

harshivcodes: Keep lifecycle logs (startup, connect, specialize trigger, exit) but remove or gate per-message logs behind a debug/verbose flag - or anything inside of handleHostMessage

gavin-aguiar (approved):

proxy/main.go
ahmedmuhsin: After specialization, the hot path (handleHostMessage with childReady=true and the bridge goroutine) has no per-message logging, only error logs. The logs in handleHostMessage's placeholder switch (WorkerInitRequest, FERR, metadata) fire once each during startup, not per request. The specializing wait log only fires during the brief window between FERR and child init. No changes needed since there are no logs in the steady-state request path.

ahmedmuhsin: See reply above. The current logs are all lifecycle logs (startup, connect, specialize, exit). There are no per-message logs in the steady-state request path after specialization.

PR #28, opened 2026-04-23
fix: proxy handles missing app binary gracefully during specialization

ahmedmuhsin:
## Problem

When a function app is first created on flex consumption, the container is specialized before any code is deployed. The proxy would crash in two scenarios:

1. **Post-specialization startup** (`WEBSITE_PLACEHOLDER_MODE=0`): proxy fatals with `Proxy requires WEBSITE_PLACEHOLDER_MODE=1`, causing a tight crash loop where the host repeatedly restarts the script host (`ConsecutiveErrors=0,1,2,...`).

2. **During FERR handling**: proxy fatals with `Failed to start child worker: fork/exec app: no such file or directory` when trying to exec the non-existent binary.

Additionally:
- When a user's Go app binary crashes, stderr output (panic stack traces, error messages) was lost -- App Insights showed bare exceptions with no inner detail (133 occurrences observed).
- When the app binary lacks execute permissions (common when deploying from Windows), the proxy crashed with a bare `exited with code 1` instead of reporting the actual issue.

### 1. Graceful handling of missing binary
Both crash scenarios are now handled gracefully:
- The proxy runs as a **no-op worker**, responding to host messages (init, heartbeats, status, metadata) with empty/success responses
- The host stays healthy with zero functions until the platform recycles pods with deployed content
- Production flow (from Kudu Legion source): after user deploys, Kudu calls `removeworker/allStandard` to kill all worker pods. New pods start fresh with content already mounted via FUSE.

### 2. Stderr capture for crash diagnostics
- Child stderr is captured into a buffer (while still piping to os.Stderr)
- On crash, captured stderr is logged via `log.Printf` so it appears in FunctionAppLogs through the host's `OnOutputDataReceived` handler
- Brief sleep before `os.Exit()` gives the host time to read the output

### 3. Executable permission check
- At startup (exec bypass): checks execute bit before `syscall.Exec`, fatals with clear message mentioning `chmod +x`
- During `specialize()`: checks execute bit before `cmd.Start()`, sends FERR failure response with `RpcException` containing the permission error -- proxy stays alive, host sees the error in App Insights

## Changes

- **proxy/main.go**:
  - Removed fatal at startup when not in placeholder mode without binary
  - Added binary existence check in `specialize()` -- responds with success FERR and enters no-op mode
  - Added executable permission check in both exec bypass and `specialize()`
  - Capture child stderr in buffer, log on crash for App Insights visibility
- **tests/consumption/specialize_before_deploy_test.go**: Three tests:
  - `TestSpecializeBeforeDeploy`: Post-specialization without binary -- verifies no crash loop
  - `TestSpecializeBeforeDeployFullFlow`: Full 4-phase lifecycle -- specialize without binary, deploy, container restart (pod recycle), specialize with binary, `/api/hello` returns 200
  - `TestSpecializeWithNonExecutableBinary`: Deploys binary without execute permissions -- verifies clear error instead of bare crash
- **tests/consumption/consumption_helpers_test.go**: Added `buildAndStartFlexSpecialized()`, `sendRequestWithTimeout()`, `restartHost()`, `restartContainer()` helpers
- **tests/consumption/Dockerfile.flex-test-specialize-before-deploy**: MCR native1.0 image with local proxy build

## Testing

All tests pass:
- `TestSpecializeBeforeDeploy` -- PASS (55s, no crash loop)
- `TestSpecializeBeforeDeployFullFlow` -- PASS (58s, full lifecycle)
- `TestSpecializeWithNonExecutableBinary` -- PASS (37s, permission error caught)
- `TestFlexConsumptionHttpTriggerProxy` -- PASS (42s, no regression)

### DIFF
--- a/proxy/main.go
+++ b/proxy/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"io"
@@ -11,6 +12,7 @@ import (
 	"path/filepath"
 	"sync"
 	"syscall"
+	"time"
 
 	"github.com/azure/azure-functions-golang-worker/worker"
 	pb "github.com/azure/azure-functions-golang-worker/worker/proto"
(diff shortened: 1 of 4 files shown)


proxy/main.go
```diff
@@ -316,6 +325,48 @@ func (p *Proxy) specialize(req *pb.FunctionEnvironmentReloadRequest) {
...
+				FunctionEnvironmentReloadResponse: &pb.FunctionEnvironmentReloadResponse{
+					Result: &pb.StatusResult{Status: pb.StatusResult_Success},
+				},
+			},
+		})
+		// Clear isSpecializing so subsequent messages (FunctionsMetadataRequest,
+		// heartbeats, etc.) are handled by the placeholder mode switch.
+		p.mutex.Lock()
+		p.isSpecializing = false
+		p.mutex.Unlock()
```
gavin-aguiar, line 354: When the binary is missing during specialize(), this resets isSpecializing = false but never closes childInitDone. This will block any goroutine. Add a   close(p.childInitDone)  after this line,

proxy/main.go
```diff
@@ -316,6 +325,48 @@ func (p *Proxy) specialize(req *pb.FunctionEnvironmentReloadRequest) {
...
+					Result: &pb.StatusResult{
+						Status:    pb.StatusResult_Failure,
+						Exception: &pb.RpcException{Message: errMsg},
+					},
+				},
+			},
+		})
+		p.mutex.Lock()
+		p.isSpecializing = false
+		p.mutex.Unlock()
```
gavin-aguiar, line 374: Same here. add  close(p.childInitDone)

proxy/main.go
ahmedmuhsin: Good catch. Instead of closing childInitDone (which would cause the blocked goroutine to try sending on a nil childStream), I added a dedicated `noChildMode` channel. The specializing block now uses `select` on both channels, `childInitDone` for the normal path, `noChildMode` to fall through to the placeholder switch. Fixed in 71d515f.

proxy/main.go
ahmedmuhsin: Fixed, same approach with the `noChildMode` channel. See 71d515f.

gavin-aguiar (approved):

PR #31, opened 2026-05-04
Add HTTP streaming support via HttpUri capability

ahmedmuhsin:
# Add HTTP streaming support via `HttpUri` capability

## Summary

Enables true end-to-end HTTP streaming for the Go worker by adopting the `HttpUri` capability used by the Python and .NET-isolated workers. User handlers now run against the live `http.ResponseWriter` from `net/http` — so `http.Flusher`, chunked transfer encoding, and streaming responses all work natively.

As a corollary, **all** HTTP-trigger traffic (streaming or not) now skips the gRPC body double-hop, matching the streaming-capable worker SKUs. Backward compatible with hosts that don't recognize `HttpUri`.

## Problem

The custom `ResponseWriterProxy` wraps a `*bytes.Buffer` and does not implement `http.Flusher`, so `flusher, ok := w.(http.Flusher)` always fails. Every `Write()` is silently buffered and the entire response is sent in one batch after the handler returns. Server-Sent Events, AI streaming responses, and any other incremental delivery is broken.

In addition, every HTTP request body is protobuf-serialized into `RpcHttp.Body` and traverses gRPC four times (host ↔ proxy ↔ user app, in both directions).

## How `HttpUri` works

1. Worker advertises `HttpUri: http://127.0.0.1:<random-port>` and `RequiresRouteParameters: true` in `WorkerInitResponse.capabilities`.
2. Host stores the URL and uses YARP (`Yarp.ReverseProxy.Forwarder`) to forward each HTTP-trigger request directly to that local URL, untouched apart from an `x-ms-invocation-id` correlation header. The gRPC `RpcHttp` becomes a near-empty notification.
3. User handler runs on the HTTP goroutine inside the embedded `http.Server`, against the live `ResponseWriter` from `net/http`.
4. Worker reports a minimal `InvocationResponse` (status only) over gRPC once the handler returns.

The HTTP and gRPC arrivals are joined by invocation id in a small rendezvous coordinator: a `sync.Mutex`-guarded `map[string]*pendingHTTPInvocation`, where each entry holds two buffered-size-1 channels (one carrying `*LoadedFunction` from gRPC to HTTP, one carrying `*pb.StatusResult` back). Whichever side arrives first waits for the other; ordering is irrelevant.

## Why this is a good fit for Go

- **Native `net/http`.** Standard library `ResponseWriter` already implements `http.Flusher`, supports chunked transfer encoding, hijacking, and trailers. No custom buffering layer.
- **No proto/host changes.** Host already supports `HttpUri`. Just opt in via a capability string.
- **No sidecar.** The HTTP listener is in-process on a loopback port — same binary, same goroutines.

### What this also gives non-streaming HTTP triggers

Once the embedded HTTP server is up, every HTTP-trigger invocation skips the gRPC body double-hop because the host stops populating `RpcHttp` for proxying workers. This means:

- No protobuf serialization of request/response bodies.
- No relay through the proxy worker for body traffic (in flex consumption).
- Native HTTP semantics — large uploads stream through without buffering, headers preserved verbatim.
- Smaller gRPC messages, less head-of-line blocking.

The only added cost is one loopback TCP hop within the container, which is negligible.

## Backward compatibility

Three layered fallbacks make this safe to ship:

1. **Old host (no `HttpUri` support).** Host ignores the capability and keeps sending the full body via gRPC. Worker detects a populated `RpcHttp.Url` (`isHTTPProxiedInvocation` returns false) and uses the existing buffered `ResponseWriterProxy` path.
2. **Listener fails to open.** `startHTTPProxy` logs a warning, returns nil, capability isn't advertised; worker continues on the gRPC body path.
3. **App has no HTTP triggers.** `appHasHTTPFunctions` returns false; the embedded server is never started. Zero cost for timer-only / queue-only / blob-only apps.

## Per-invocation resilience

Pathological per-invocation conditions never crash the worker. User-handler panics are recovered and surfaced as a `Failure` `InvocationResponse`. Client disconnects (YARP cancels the forward) push a Failure status into the rendezvous so the gRPC side unblocks immediately rather than waiting on the 4-minute hard timeout. Rendezvous timeouts themselves return a `Failure` status to the host instead of a Go error — the worker stays up to serve subsequent invocations.

## Opt-out: `FUNCTIONS_GO_DISABLE_HTTP_PROXY`

Per-app environment variable to force the legacy gRPC body path even when the host supports `HttpUri`:

```
FUNCTIONS_GO_DISABLE_HTTP_PROXY=1
```

Accepts `1`, `true` (case-insensitive), or any non-empty non-zero value. Logged at startup so it's visible in App Insights:

```
HTTP proxy: disabled via FUNCTIONS_GO_DISABLE_HTTP_PROXY=1, using gRPC body for HTTP triggers
```

Use cases: bug isolation / panic switch (flip the env var to confirm whether a regression is in the new HTTP path without redeploying); telemetry middleware that reads bodies from `RpcHttp.Body`; restricted sandbox configs that block loopback listeners; custom gRPC interceptors / service-mesh setups.

There is no opt-*in* counterpart. Streaming is automatic when the worker has HTTP triggers, mirroring Python and .NET-isolated.

## Changes

- **`worker/http_proxy.go`** (new). Embedded `http.Server` on a random loopback port, plus the invocation rendezvous coordinator. `handle` recovers from panics so the gRPC side always gets a response.
- **`worker/handlers.go`**. `handleWorkerInitRequest` advertises the standard out-of-proc capability set plus `HttpUri` and `RequiresRouteParameters` when the proxy is up. `handleInvocationRequest` short-circuits HTTP-trigger invocations through the coordinator when proxying is in effect, surfacing rendezvous failures as `Failure` `InvocationResponse`s rather than fatal Go errors.
- **`worker/worker.go`, `worker/dispatcher.go`**. Lifecycle wiring.
- **`worker/http_proxy_test.go`** (new). Unit tests including end-to-end streaming, panicking handler, env-var truth table, no-HTTP-functions short-circuit.
- **`tests/consumption/http_streaming_test.go`** (new). Two Docker-based integration tests against the real flex consumption MCR base image:
  - `TestFlexConsumptionHttpStreaming` — deploys the `httpStreaming` sample, asserts 5 SSE events arrive with `>2s` spread between first and last (handler emits at 1s intervals; buffered responses arrive in <1s, streamed responses arrive over ~4s). Last run: 4.00s spread.
  - `TestFlexConsumptionHttpStreamingOptOut` — deploys the `httpTrigger` sample with `FUNCTIONS_GO_DISABLE_HTTP_PROXY=1` in the encrypted assign context. Asserts the opt-out log line appears in container logs and `/api/hello` returns 200 via the gRPC path.

## How to verify

```pwsh
# Unit tests (fast)
go test ./worker/...

# Integration tests (Docker required, ~95s for the four HTTP-related tests)
cd tests/consumption
go test -run "TestFlexConsumptionPlaceholderPing|TestFlexConsumptionHttpTriggerProxy|TestFlexConsumptionHttpStreaming$|TestFlexConsumptionHttpStreamingOptOut" -v -timeout 30m
```

## Risks

- **Header pass-through.** YARP forwards almost everything verbatim, but if user code depended on the gRPC-shape headers (`X-Forwarded-*`, `X-Original-Url`) being precisely what was in `RpcHttp`, they may see tiny shape differences. Opt-out is available.
- **Telemetry.** Observability tools that read request bodies from `RpcHttp.Body` will see empty bodies on the new path. Documented in the opt-out section.

### DIFF
--- a/worker/http_proxy.go
+++ b/worker/http_proxy.go
@@ -1,0 +1,336 @@
+package worker
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"os"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/azure/azure-functions-golang-worker/sdk"
+	pb "github.com/azure/azure-functions-golang-worker/worker/proto"
+)
+
+// HTTP proxying / streaming.
+//
+// When the user app contains HTTP triggers, the worker spawns a small
+// http.Server on a loopback port and advertises that URL to the host via the
+// "HttpUri" capability in WorkerInitResponse. The host then forwards every
+// incoming HTTP request directly to that local server (using YARP), with the
+// gRPC InvocationRequest carrying only correlation metadata.
+//
+// This means the user's net/http handler runs against the *real* live
+// http.ResponseWriter from net/http, so http.Flusher, chunked transfer
+// encoding, trailers, and streaming bodies all work natively — no buffering
+// proxy in the worker.
+//
+// The HTTP and gRPC arrivals are coordinated by invocation id (header
+// "x-ms-invocation-id"). Whichever side arrives first waits in the
+// coordinator for the other; the user handler runs on the HTTP goroutine,
+// and the result is shipped back to the gRPC goroutine to populate the
+// minimal InvocationResponse.
+
+// invocationCorrelationHeader is the header the host adds to forwarded
+// requests so the worker can correlate an HTTP request with its
+// gRPC InvocationRequest. Matches ScriptConstants.HttpProxyCorrelationHeader
+// in the Functions host.
+const invocationCorrelationHeader = "x-ms-invocation-id"
+
+// httpInvocationWaitTimeout caps how long either side waits for the other.
+// The host's reverse-proxy ActivityTimeout is 240s, so we stay well under
+// that for the join, but the HTTP handler itself can run as long as the
+// underlying request stays open.
+const httpInvocationWaitTimeout = 4 * time.Minute
+
+// pendingHTTPInvocation tracks the two-arrival rendezvous between the gRPC
+// invocation goroutine and the HTTP request goroutine for a single
+// invocation id.
+//
+// Buffered channels of size 1 mean that whichever side arrives first never
+// blocks on the send, so order of arrival is irrelevant.
+type pendingHTTPInvocation struct {
+	grpc chan *grpcInvocationSide // delivered by gRPC dispatcher
+	done chan *grpcInvocationResult // delivered by HTTP handler when finished
+}
+
+// grpcInvocationSide is what the gRPC dispatcher contributes to the
+// rendezvous: the function identity and any host-supplied trigger metadata.
+type grpcInvocationSide struct {
+	functionID  string
+	loadedFunc  *LoadedFunction
+	triggerMeta map[string]*pb.TypedData
+	inputData   []*pb.ParameterBinding
+}
+
+// grpcInvocationResult is what the HTTP side reports back so the gRPC
+// dispatcher can produce a minimal InvocationResponse.
+type grpcInvocationResult struct {
+	status *pb.StatusResult
+}
+
+// httpProxy is the in-process HTTP server that the host forwards trigger
+// requests to. There is at most one per worker process.
+type httpProxy struct {
+	url      string
+	server   *http.Server
+	listener net.Listener
+
+	mu      sync.Mutex
+	pending map[string]*pendingHTTPInvocation // keyed by invocation id
+}
+
+// disableHTTPProxyEnvVar lets users force the worker back onto the legacy
+// gRPC-body HTTP path even when the host supports HttpUri proxying. Set to
+// "1", "true", or any non-empty value to disable the embedded HTTP server.
+//
+// Use cases:
+//   - Debugging: confirming a regression is/isn't caused by the HTTP-proxy path.
+//   - Telemetry tooling that observes request bodies via the gRPC RpcHttp
+//     message and would otherwise see an empty body when proxying is on.
+//   - Restricted environments where opening a loopback listener is not allowed
+//     (forces the fallback explicitly instead of relying on listener failure).
+const disableHTTPProxyEnvVar = "FUNCTIONS_GO_DISABLE_HTTP_PROXY"
+
+// startHTTPProxy starts the loopback HTTP server. It returns nil (and logs)
+// if the user app registered no HTTP functions, the user disabled HTTP
+// proxying via FUNCTIONS_GO_DISABLE_HTTP_PROXY, or the listener could not
+// be opened — in any of these cases the worker falls back to the
+// gRPC-buffered HTTP path.
+func startHTTPProxy(app *sdk.App) *httpProxy {
+	if v := os.Getenv(disableHTTPProxyEnvVar); v != "" && v != "0" && !strings.EqualFold(v, "false") {
+		log.Printf("HTTP proxy: disabled via %s=%s, using gRPC body for HTTP triggers", disableHTTPProxyEnvVar, v)
+		return nil
+	}
+
+	if !appHasHTTPFunctions(app) {
+		return nil
+	}
+
+	lis, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		log.Printf("HTTP proxy: failed to open loopback listener, falling back to gRPC body: %v", err)
+		return nil
+	}
+
+	p := &httpProxy{
+		listener: lis,
+		pending:  make(map[string]*pendingHTTPInvocation),
+		url:      fmt.Sprintf("http://%s", lis.Addr().String()),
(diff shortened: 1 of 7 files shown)


worker/http_proxy.go
```diff
@@ -0,0 +1,337 @@
...
+		pending.done <- &pb.StatusResult{
+			Status: pb.StatusResult_Failure,
+			Exception: &pb.RpcException{
+				Message: "client disconnected before invocation could start",
+				Source:  "HTTP proxy",
+			},
+		}
+		p.deletePending(invocationID)
+		return
+	case <-time.After(httpInvocationWaitTimeout):
```
gavin-aguiar: time.After in the select leaks a timer per invocation until it fires. For a 4-minute duration this means each active invocation holds a goroutine+timer for the full timeout even after it's been handled. Use time.NewTimer + defer t.Stop() for the HTTP-side wait.

worker/worker.go
```diff
@@ -15,6 +15,12 @@ func Start(app *sdk.App) {
 
 	dispatcher := NewDispatcher(config, app)
 
+	// Start the in-process HTTP proxy (used for HTTP streaming via the
+	// "HttpUri" capability). Returns nil if the app has no HTTP triggers
+	// or the loopback listener can't be opened — in either case the worker
+	// falls back to the gRPC-buffered HTTP path.
+	dispatcher.HTTPProxy = startHTTPProxy(app)
```
harshivcodes, line 22: https://github.com/Azure/azure-functions-golang-worker/blob/main/worker/grpc.go#L60

This PR assumes that the gRPC channel is capable of handling multiple invocation requests sequentially — that is, that the existing single-goroutine handleBidiStream loop
in worker/grpc.go can keep pulling messages off the wire fast enough that any per-request blocking is negligible. However, the new HTTP-streaming path fundamentally
changes the duration of one "sequential step."

Pre-PR, process(InvocationRequest) ran the user's handler inline and returned in milliseconds. Sequential dispatch was effectively concurrent in practice because the
per-message wait was tiny.

Post-PR, process(InvocationRequest) for any HTTP trigger calls notifyGRPCArrival, which blocks on <-pending.done until the user's HTTP handler returns. For SSE, LLM token
streams, log tails, or any long-poll endpoint that's seconds to minutes — sometimes indefinite. The single dispatcher goroutine is now parked for the full handler
duration.

The result: the assumption "we can drain the gRPC stream sequentially" silently changes from "always true" to "never true for the workloads streaming was built to serve."
While one streaming handler runs:

 - The next InvocationRequest sits unread in the gRPC receive buffer — the second user's HTTP body has already arrived at the loopback server and is parked in the 
rendezvous waiting for a pending.grpc send that won't happen.
 - WorkerStatusRequest health pings queue behind it; the host's status timeout fires and recycles the worker mid-stream.
 - WorkerTerminate and FunctionEnvironmentReloadRequest queue too, breaking graceful drain on deployments.

Effective per-worker HTTP concurrency drops to 1, regardless of MaxConcurrentRequests in host.json. The streaming feature ships its primary use case — many concurrent
slow clients — into a bottleneck of one.

The fix is to remove the assumption rather than try to satisfy it: spawn each InvocationRequest on its own goroutine inside handleBidiStream and serialize only 
client.Send with a sync.Mutex (gRPC ClientStream.SendMsg is not safe for concurrent use; Recv stays single-goroutine by design). Python and .NET-isolated workers already
do this for exactly this reason.

Image

harshivcodes: Image

I had copilot repro the bug locally

harshivcodes (changes requested):

worker/worker.go
ahmedmuhsin: Good catch. You're right that the single-goroutine `handleBidiStream` is incompatible with the streaming contract. Pushed a fix.

`worker/grpc.go`: every gRPC message now runs on its own goroutine, with a `sync.Mutex` serializing `client.Send`. Worker-side ordering is irrelevant since the host correlates responses by `function_id` / `invocation_id` / `request_id`, and control-plane sequencing is the host's job. Matches the dispatch model used by Python and .NET-isolated.

`tests/consumption/http_streaming_test.go`: added `TestFlexConsumptionHttpStreamingConcurrent`. Two parallel `/api/stream` requests; asserts the second's first event arrives before the first's last event. Last run:

```
A.event[0] at +10ms,  A.event[4] at +4.02s
B.event[0] at +260ms, B.event[4] at +4.26s
```

Fully overlapped, ~4.26s wall vs ~10s if serialized. All 5 integration tests pass. PTAL.

gavin-aguiar (approved):

worker/http_proxy.go
ahmedmuhsin: Good catch, fixed. Replaced `time.After` with `time.NewTimer` + `defer t.Stop()` in both rendezvous sites — the HTTP-side wait in `handle` and the gRPC-side wait in `notifyGRPCArrival`. All 5 integration tests pass.

gavin-aguiar (approved):

harshivcodes (approved):

PR #32, opened 2026-05-05
Add InvocationContext, middleware seam, slog handler, and otelfunc OpenTelemetry middleware

ahmedmuhsin:
## Summary

This PR adds three foundational capabilities to the Go worker and ships a production-ready OpenTelemetry middleware built on top of them:

1. **`sdk.InvocationContext`** — per-invocation metadata (invocation id, function name, trigger type, W3C trace context, baggage, retry info, outbound trace attributes) carried on `context.Context` and reachable via `sdk.FromContext(ctx)`.
2. **`sdk.Middleware` seam** — a small interface (`Wrap(next Handler) Handler`) plus optional `CapabilityProvider` and `ShutdownProvider` contracts that lets any user or third-party module decorate every invocation with cross-cutting behavior (tracing, structured logging, auth, retry policies, panic recovery, etc.).
3. **Structured logging via `slog`** — the worker installs an `slog`-routing handler at package init that emits every record as a typed `RpcLog` on the gRPC stream, automatically attaching `invocation_id`, `function_name`, and `trigger_type` to each entry. The worker-internal pipeline (writer + handlers + bridge observer) lives in a dedicated `worker/log/` package.
4. **`middleware/otelfunc`** — an opt-in OpenTelemetry middleware that wires distributed tracing, structured-log → OTel-log bridging, W3C trace-context propagation, inbound baggage hydration, force-flush on consumption-style plans, graceful shutdown, and zero-config setup from the standard `OTEL_EXPORTER_OTLP_*` env vars.

The OTel middleware was validated end-to-end on a Flex Consumption Function App exporting OTLP/HTTP to New Relic — worker spans, user log records, host AspNetCore spans, host init activity, and outbound `otelhttp` client spans all correlate by `trace.id`.

It is **truly opt-in**: users who don't import `middleware/otelfunc` get zero OpenTelemetry packages compiled into their binary (measured: 17.87 MB vs. 21.25 MB).

## Motivation

The pre-existing worker had three structural gaps that this PR addresses together:

- **No way to access invocation metadata from user code.** Handlers received only the trigger-typed data (`bindings.TimerInfo`, `*blob.Client`, etc.) — invocation id, function name, trace context, and retry state were locked inside the dispatcher.
- **No extensibility point for cross-cutting behavior.** Users couldn't add tracing, structured logging, auth, or any other decoration around invocations without forking the worker.
- **No structured-logging path.** `log.Printf` calls landed on stderr where the host treated them as opaque `Host.Function.Console` traces in Application Insights — no severity, no structured attributes, no per-record category.

Solving all three with one design — `InvocationContext` on `context.Context`, a Middleware interface that takes the same `context.Context`, and an `slog` handler that reads the InvocationContext off `context.Context` — keeps the worker's user surface small (no new mandatory parameters on handler signatures) and lets the OTel middleware be a regular consumer of the same APIs everyone else uses, with no privileged path through the worker.

### 1. `sdk.InvocationContext` carried on `context.Context`

A new `InvocationContext` struct exposes everything the host sends us:

```go
type InvocationContext struct {
    InvocationID    string
    FunctionID      string
    FunctionName    string
    TriggerType     string
    TraceContext    TraceContext  // TraceParent, TraceState, Attributes, Baggage
    RetryContext    RetryContext  // RetryCount, MaxRetryCount
    TriggerMetadata map[string]string
}
```

User code retrieves it the standard Go way:

```go
func TimerHandler(ctx context.Context, timer bindings.TimerInfo) error {
    ic, _ := sdk.FromContext(ctx)
    if ic != nil && ic.RetryContext.RetryCount > 0 {
        slog.WarnContext(ctx, "running on retry",
            "retry_count", ic.RetryContext.RetryCount,
            "max_retry_count", ic.RetryContext.MaxRetryCount)
    }
    return nil
}
```

For HTTP handlers, it's reachable via `r.Context()`:

```go
func HTTPHandler(w http.ResponseWriter, r *http.Request) {
    ic, _ := sdk.FromContext(r.Context())
    // ...
}
```

The dispatcher populates the InvocationContext from `pb.InvocationRequest` once per invocation and attaches it via `sdk.NewContext` before the middleware chain runs.

#### Framework-side: `MiddlewareContext`

User-facing handler signatures (`func(w http.ResponseWriter, r *http.Request)`, `func(ctx context.Context, timer bindings.TimerInfo) error`, etc.) are unchanged — user code reaches per-invocation metadata via `sdk.FromContext(ctx)` and gets back an `*InvocationContext`. The middleware chain, however, composes around a slightly different shape: every `Handler` receives an `*sdk.MiddlewareContext` that embeds `*InvocationContext` so all the trigger-side fields are still reachable by field promotion, but adds framework-only methods that the worker dispatcher reads after the chain returns. The most important one is `SetOutboundTraceAttribute(key, value)`, whose values are propagated onto the host's parent AspNetCore span via `InvocationResponse.TraceContextAttributes` for **both** the gRPC-body invocation path and the HTTP-streaming proxy path.

User code does not normally call `SetOutboundTraceAttribute` directly. The idiomatic path is to call `span.SetAttributes(...)` on the worker invocation span; the `otelfunc` middleware auto-harvests those attributes at end-of-invocation and records them on the `MiddlewareContext` for you (see [middleware/otelfunc/harvest.go](middleware/otelfunc/harvest.go)). Middleware authors that need direct control — e.g. an auth middleware that wants to tag the host span with `tenant_id` resolved before any user span exists — can reach the `MiddlewareContext` via `sdk.MiddlewareContextFrom(ctx)` and call the setter.

### 2. The `sdk.Middleware` seam

A small interface plus optional contracts:

```go
// Middleware decorates a Handler with cross-cutting behavior.
type Middleware interface {
    Wrap(next Handler) Handler
}

// MiddlewareFunc adapts a plain function to the Middleware interface
// — the same Handler/HandlerFunc pattern as net/http.
type MiddlewareFunc func(next Handler) Handler

// CapabilityProvider lets a Middleware advertise worker-level
// capability flags to the host (e.g. WorkerOpenTelemetryEnabled).
type CapabilityProvider interface {
    Capabilities() map[string]string
}

// ShutdownProvider lets a Middleware register cleanup callbacks
// for asynchronous resources (exporters, batch processors, etc.).
type ShutdownProvider interface {
    Shutdown(ctx context.Context) error
}
```

Users register middleware via `App.Use(mw)`. Registration order matches the standard Go convention: first-registered is outermost (runs first on entry, last on exit). Middleware that also implement `CapabilityProvider` have their capability map merged into `WorkerInitResponse.Capabilities`; middleware that implement `ShutdownProvider` get a deterministic, bounded cleanup on graceful shutdown (gRPC stream closing or SIGTERM/SIGINT).

Sample — a tiny timing middleware shown in [samples/middleware/main.go](samples/middleware/main.go):

```go
func timingMiddleware(next sdk.Handler) sdk.Handler {
    return func(ctx context.Context, mc *sdk.MiddlewareContext) error {
        start := time.Now()
        err := next(ctx, mc)
        slog.InfoContext(ctx, "invocation finished",
            "function", mc.FunctionName, // promoted from embedded *InvocationContext
            "duration_ms", time.Since(start).Milliseconds(),
            "err", err,
        )
        return err
    }
}

func main() {
    app := sdk.FunctionApp()
    app.Use(sdk.MiddlewareFunc(timingMiddleware))
    app.HTTP("hello", HTTPTriggerHandler, sdk.WithMethods("GET"))
    worker.Start(app)
}
```

The HTTP-streaming path (host's HttpUri capability) runs through the same `App.Compose(inner)` chain as gRPC-body invocations, so registered middleware wraps every trigger type uniformly.

### 3. `slog`-based structured logging

The SDK installs an `slog` handler at package init that:

- Routes every record over the gRPC stream as a typed `RpcLog` (no more stderr `log.Printf`).
- Encodes structured attributes into `RpcLog.PropertiesMap` (the typed field the host forwards to `customDimensions` in App Insights) and renders them into `RpcLog.Message` in logfmt format so they survive transports that ignore properties.
- Automatically attaches `invocation_id`, `function_name`, and `trigger_type` to every record emitted inside an invocation context.
- Honors slog's order-sensitive `With` / `WithGroup` semantics: attributes bound BEFORE a `WithGroup` stay at the top level; attributes bound AFTER nest inside that group. Handlers track each bound attr's bind-time group stack, matching `slog.JSONHandler` / `slog.TextHandler`.
- Honors the host's `log_categories` filter map sent in `WorkerInitRequest`, with a default-allow when no filters are present (matches dotnet-isolated behavior; previously user logs emitted during cold start were silently dropped).
- Buffers records emitted before the gRPC stream is ready and flushes them on connect (`worker/log/bootstrap.go`), so logs from arg parsing or gRPC dial errors aren't lost.

Sample call from [samples/timerTrigger/main.go](samples/timerTrigger/main.go):

```go
slog.InfoContext(ctx, "timer trigger executed",
    "is_past_due", timer.IsPastDue,
    "last", timer.ScheduleStatus.Last,
    "next", timer.ScheduleStatus.Next,
)
```

The record arrives at the host with `severityLevel=Information`, properties `{is_past_due, last, next, invocation_id, function_name, trigger_type}`, and a message string that lists every attribute — readable in any backend that consumes the log message text directly.

The worker emits a one-time `Go worker started` log record on cold start summarizing the SDK version, git revision, Go version, and runtime metadata. It flows through the User-category gRPC path so it lands in NR via the otelfunc bridge when OTel is in use, and in App Insights under the function category otherwise. Customer queries against `message = "Go worker started"` show which build is running and whether it carries any local `replace` directive in `go.mod`.

A separate System-category logger is used for worker-internal events (dispatcher startup, message dispatch, function load) so they don't pick up user invocation attrs.

Worker-internal slog call sites use `slog.LogAttrs` (the typed form) instead of variadic `key, value` pairs, so a wrong type or missing value in a log attribute is a compile error rather than a silent `!BADKEY` record. CI runs the `sloglint` linter against the same rule for new code.

### 4. `middleware/otelfunc` — OpenTelemetry distributed tracing

A focused middleware in a separate package. The smallest setup is the zero-arg form combined with the standard OTel env vars on the Function App:

```go
import (
    "github.com/azure/azure-functions-golang-worker/middleware/otelfunc"
)

func main() {
    app := sdk.FunctionApp()
    app.Use(otelfunc.Middleware())
    app.HTTP("hello", helloHandler)
    worker.Start(app)
}
```

```
OTEL_EXPORTER_OTLP_ENDPOINT=https://otlp.your-backend.example
OTEL_EXPORTER_OTLP_HEADERS=api-key=<token>
OTEL_SERVICE_NAME=my-function-app
```

With that, every invocation gets:

- An **internal-kind** span named `function <FunctionName>` carrying `faas.invocation_id`, `faas.name`, `faas.trigger` semconv attributes plus `process.pid`, `faas.instance`, and `azure.functions.live_logs_session_id` promoted from the host's inbound `RpcTraceContext.Attributes`. The host's AspNetCore instrumentation owns the SERVER-kind span for HTTP triggers, and non-HTTP triggers don't represent a network server, so the worker's span is INTERNAL — matching the Java worker.
- W3C `traceparent`/`tracestate` extracted from `RpcTraceContext` so worker spans correlate with the host's parent activity.
- Inbound W3C baggage hydrated onto `ctx` via `baggage.ContextWithBaggage`, so user code calling `baggage.FromContext(ctx)` sees what upstream services attached.
- `ic.OutboundTraceAttributes` propagated onto the host's parent AspNetCore span via `InvocationResponse.TraceContextAttributes` — works on both the gRPC-body and HTTP-streaming paths.
- `ForceFlush` on both the owned TracerProvider and LoggerProvider after every invocation — critical on consumption-style plans where the container can be frozen between invocations.
- Graceful shutdown via `sdk.ShutdownProvider` when the gRPC stream closes or the process receives SIGTERM/SIGINT.
- Bidirectional log/trace correlation: user `slog` records emitted during the invocation carry the same `trace.id`/`span.id` as the worker span the middleware created.

#### Resolution priority

The middleware picks where the TracerProvider (and separately, LoggerProvider) comes from in this order:

1. `WithTracerProvider(tp)` / `WithLoggerProvider(lp)` — explicit, use as-is.
2. `WithExporter(e)` / `WithLogExporter(e)` — caller hands exporters; the middleware builds an owned provider with a default Resource. **Stackable**: call multiple times to fan out to multiple backends.
3. `OTEL_EXPORTER_OTLP_ENDPOINT` env var — auto-build an OTLP HTTP provider.
4. `otel.GetTracerProvider()` / `global.GetLoggerProvider()` — honor a non-noop global.

Auto-OTLP intentionally wins over the global. "Auto-OTLP" is what we call the middleware's path of building its own OTLP HTTP TracerProvider / LoggerProvider from the standard `OTEL_EXPORTER_OTLP_ENDPOINT` env var. The OTel global TracerProvider (`otel.GetTracerProvider()`) is **not** the noop default by reference any more — it's a delegating wrapper that routes spans through `go.opentelemetry.io/auto/sdk`, which is pulled in transitively by `otelslog` and several other OTel contrib bridges. That auto SDK is designed for an eBPF agent to attach later: it reports spans as `IsRecording=true` even when nothing is actually exporting, so a behavioral noop check ("start a span, ask `IsRecording`") returns false ("not noop") whenever any contrib bridge is imported.

If we treated the wrapper as "user already configured a TracerProvider" and skipped auto-OTLP, the common case — user sets `OTEL_EXPORTER_OTLP_ENDPOINT` and calls `app.Use(otelfunc.Middleware())` — would silently swallow every span. So when nothing explicit is supplied via `WithTracerProvider`/`WithExporter`, we check the env var **before** the global: if the env var is set, we build our own OTLP provider; only if it isn't do we fall back to whatever the global delegates to. The same reasoning applies to `global.GetLoggerProvider()` for the log path.

#### Default Resource and `WithResource`

When the middleware builds its own provider, the Resource carries the same set the Java worker emits (so cross-runtime dashboards work with one set of queries):

- `cloud.provider = azure`
- `cloud.platform = azure_functions`
- `cloud.region` (from `REGION_NAME`, when set)
- `cloud.resource_id = /subscriptions/<sub>/resourceGroups/<rg>/providers/Microsoft.Web/sites/<site>` (built from `WEBSITE_OWNER_NAME` + `WEBSITE_RESOURCE_GROUP` + `WEBSITE_SITE_NAME`)
- `deployment.environment.name` (from `WEBSITE_SLOT_NAME`, default `production`)
- `service.name` (from `OTEL_SERVICE_NAME` → `WEBSITE_SITE_NAME` → `"azure-functions"`)
- `otel.library.version` set to the SDK module's runtime version, so `otel.library.version = ...` filters telemetry by the exact SDK build

Note: `cloud.resource_id` is omitted on Flex Consumption because the platform doesn't inject `WEBSITE_RESOURCE_GROUP`. The .NET host's `FunctionsResourceDetector` has the same gap. Customers can supply the value via `OTEL_RESOURCE_ATTRIBUTES` or `WithResource(...)`.

Extend the Resource with typed deployment metadata via `WithResource(...)` — useful for values that aren't suitable for `OTEL_RESOURCE_ATTRIBUTES` (typed numerics, `-ldflags`-injected build SHAs, etc.):

```go
app.Use(otelfunc.Middleware(
    otelfunc.WithResource(
        semconv.ServiceVersion(buildVersion),
        semconv.DeploymentEnvironmentName("production"),
        attribute.String("build.sha", buildSHA),
    ),
))
```

#### Multi-backend fan-out

`WithExporter` and `WithLogExporter` are stackable so users can in-process fan out to several backends without constructing a custom TracerProvider:

```go
otlpExp, _ := otlptracehttp.New(ctx)
debugExp, _ := stdouttrace.New(stdouttrace.WithWriter(os.Stderr))

app.Use(otelfunc.Middleware(
    otelfunc.WithExporter(otlpExp),
    otelfunc.WithExporter(debugExp),
))
```

For multi-backend routing without changing user code, a sidecar OpenTelemetry Collector (point `OTEL_EXPORTER_OTLP_ENDPOINT` at it) is the operator-side option.

#### Capability advertising

When a non-noop TracerProvider is in play and the kill switch is not set, the middleware advertises:

- `WorkerOpenTelemetryEnabled = "true"`
- `WorkerOpenTelemetrySchemaVersion = "1.37.0"` (must match the imported `go.opentelemetry.io/otel/semconv/vX.Y.Z` version)

via `sdk.CapabilityProvider`. The narrow effect of this capability (verified against the host's `OpenTelemetryConfigurationExtensions`) is that the host suppresses forwarding of worker-emitted `Function.*` log records into its own OpenTelemetry log pipeline — the worker is expected to emit those itself via its own LoggerProvider. Host AspNetCore HTTP-server spans, the Microsoft.Azure.Functions.Host init activity, and `Host.*` traces are unaffected (they describe host activity, not worker activity).

#### Kill switch

Set `AZURE_FUNCTIONS_WORKER_OPENTELEMETRY_DISABLED=true` to make the middleware a pass-through without redeploying — useful for triage and per-slot rollouts.

### 5. The OTel-opt-in design

The `worker` package never imports any OTel module. The slog→OTel bridge lives in `middleware/otelfunc` and is wired through a hook on the worker's logging package:

```go
// In worker/log/observer.go (no OTel imports here)
type Observer func(ctx context.Context, record slog.Record)
func RegisterObserver(fn Observer)
```

`otelfunc.Middleware()` registers exactly one observer (via `sync.Once`) that bridges every user `slog` record through `otelslog` to the configured `LoggerProvider`. Users who don't import `middleware/otelfunc` never link in any OTel code:

| Configuration | OTel packages compiled in | Binary size |
|---|---:|---:|
| `sdk` + `worker` only (no otelfunc) | **0** | 17.87 MB |
| `+ app.Use(otelfunc.Middleware())` | 61 | 21.25 MB |

## HTTP streaming integration

The HTTP-streaming feature from #31 is preserved end-to-end. Two fixes during this PR's review pass:

1. **Middleware on the streaming path.** HTTP-streaming invocations were initially bypassing `App.Compose` entirely — `otelfunc.Middleware` (and any other middleware) would have silently skipped streaming HTTP triggers. The fix threads `*sdk.App` and `*pb.InvocationRequest` through the proxy's rendezvous channel and runs the user handler inside `app.Compose(inner)`. `TestHTTPProxy_RunsMiddlewareChain` locks the contract in.

2. **Outbound trace attribute round-trip.** The streaming path's rendezvous channel originally carried only `*pb.StatusResult`, so anything middleware (or auto-harvest) recorded on the per-invocation context inside a streaming handler was silently dropped before reaching `InvocationResponse.TraceContextAttributes`. The channel now carries `*sdk.MiddlewareContext` (the same `mc` the gRPC-body path runs the chain with), so the streaming-path response builder reads invocation-output fields off the same struct. `mc` is constructed BEFORE the panic-recover block runs, so partial mutations made before a user-handler panic still propagate — matching the gRPC-body path's semantics. `TestHTTPProxy_PropagatesIcToGRPCSide` and `TestHTTPProxy_IcSurvivesHandlerPanic` lock the behavior in.

## Validation

- All packages build clean on Go 1.24+ (verified via `go mod tidy -go=1.24.0 -compat=1.24`).
- `go test ./...` passes — ~135 unit tests across `sdk`, `worker`, `worker/log`, and `middleware/otelfunc`.
- `go vet ./...` and `golangci-lint` (with `sloglint`, `revive`, `tagliatelle`) clean on the touched packages.
- **End-to-end production validation**: multiple deployment rounds against a Flex Consumption Function App in Central US exporting OTLP/HTTP to New Relic. Each round verified worker spans, user log records (with `trace.id`/`span.id`), outbound trace-attribute propagation onto host parent activity (gRPC-body and streaming paths, auto-harvested from `span.SetAttributes`), outbound HTTP child spans via `otelhttp.NewTransport`, and host AspNetCore + Functions.Host init activity all land in the same backend correlated by trace id.
- Coverage: `middleware/otelfunc` 86%+, `sdk` 91%+, `worker` 65%+ (reflection-heavy dispatcher paths drag this down; new code is well-covered).

## Notable changes by area

| Area | Purpose |
|---|---|
| `middleware/otelfunc/` | New OTel middleware package; auto-OTLP, multi-backend fan-out, log↔trace correlation, default Resource, capability advertising, kill switch |
| `sdk/` | `InvocationContext` (user-facing), `MiddlewareContext` (framework-side, with `SetOutboundTraceAttribute`), `Middleware` seam, slog adapter (`NewLogHandler`, `SetDefaultBaseHandler`, auto-install via `init`) |
| `worker/` | RpcLog routing, capability advertising, middleware chain plumbing on both gRPC-body and HTTP-streaming paths |
| `worker/log/` | New package: bootstrap stderr handler, gRPC `Writer`, User-/System-category slog handlers, observer registry |
| `worker/proto/` | Proto regen for `WorkerMetadata`, `Baggage`, `TraceContextAttributes` |
| `samples/` | Modernized to `slog`. New focused samples: `samples/middleware/` (two middleware styles + `MiddlewareContext` usage) and `samples/otelTracing/` (host.json `telemetryMode: OpenTelemetry` + otelfunc end-to-end). `httpStreaming` demonstrates the streaming-path slog pattern. |
| `tests/integration/` | Tiny log-message updates to match new slog text |
| `README.md`, `TECHNICAL_SPEC.md`, `middleware/otelfunc/README.md` | New observability section / architecture coverage / middleware reference |

## Behavioral changes for existing users

- `log.Printf(...)` to stderr no longer appears in App Insights as `Host.Function.Console` traces by default. Migrate to `slog.InfoContext(ctx, ...)` — the sample updates demonstrate the pattern. Stdlib `log.Print*` still reaches NR via the host's stderr capture but does not carry `trace.id`/`span.id`/`invocation_id`, because `log.Print*`'s signature has no `context.Context` parameter (predates context). This is a Go stdlib limitation, not a configuration issue.
- The worker advertises new capabilities (`WorkerOpenTelemetryEnabled`, etc.) only when the user opts in via `app.Use(otelfunc.Middleware())`. Without that call the worker behaves like before from the host's perspective.

## Risks

- Concurrent dispatch model (one goroutine per inbound message, introduced together with HTTP streaming on `main`) now has explicit panic recovery in the dispatcher goroutine — a panicking user handler logs the stack via the system logger instead of crashing the worker.
- gRPC `Send` serialization is owned by a single sender goroutine (`startSender`) draining a 256-buffered channel; the stop function is `sync.Once`-guarded so repeated/concurrent stop calls cannot double-close the queue.
- `cloud.resource_id` is omitted on Flex Consumption because `WEBSITE_RESOURCE_GROUP` is not platform-injected. The .NET host has the same gap. Documented in the otelfunc package godoc and the README.
- Out-of-process behavior was production-tested on Flex Consumption in Central US against New Relic. Linux Consumption (different freeze characteristics) and gRPC-OTLP (only OTLP/HTTP tested) are unchanged from the OpenTelemetry SDK defaults and not specifically validated.

## Related work

- Builds on top of the HTTP-streaming PR #31 (HttpUri capability, per-message goroutine dispatch).
- The proto regen pulls in new fields used by this work (`WorkerMetadata`, `RpcTraceContext.Baggage`, `InvocationResponse.TraceContextAttributes`) plus an `AppCapabilities` field reserved by the host that we don't currently emit.


worker/handlers.go
```diff
@@ -259,14 +279,32 @@ func handleInvocationRequest(req *pb.InvocationRequest, disp *Dispatcher, reques
...
 
+	// Build the per-invocation context that user handlers and middleware will
+	// see. Stash it on context.Context — every handler that takes a
+	// context.Context can retrieve it via sdk.FromContext.
+	ic := buildInvocationContext(req, &loadedFunc.Function)
+	ctx := sdk.NewContext(context.Background(), ic)
+
+	// Emit a system log with the inbound trace_parent so OTel correlation
+	// can be debugged from production logs without instrumenting user code.
+	disp.SystemLogger().Debug("InvocationRequest trace context",
```
harshivcodes: We should enable sloglint in CI - if key value pairs are not correctly provided, this code doesn't throw an error but it silently logs, with a !BADKEY attribute for the missing key.

We should use LogAttrs(ctx, level, ...) wherever feasible.
```
disp.SystemLogger().LogAttrs(ctx, slog.LevelDebug, "InvocationRequest trace context",
     slog.String("invocation_id", req.GetInvocationId()),
     slog.String("trace_parent", tc.GetTraceParent()),
     slog.String("trace_state", tc.GetTraceState()),
     slog.Int("baggage_count", len(tc.GetBaggage())),
 )
```

sdk/log.go
```diff
@@ -0,0 +1,172 @@
...
+func (h *invocationLogHandler) resolveBase() slog.Handler {
+	if h.explicitBase != nil {
+		return h.explicitBase
+	}
+	return defaultBase()
+}
+
+// effective composes the resolved base with any accumulated groups/attrs
+// for use in Handle.
+func (h *invocationLogHandler) effective() slog.Handler {
```
harshivcodes: Maybe nit; all three custom slog handlers — invocationLogHandler in sdk/log.go, and userLogHandler + systemLogHandler in worker/system_logger.go — store bound attributes and groups in flat parallel slices and then apply all groups before all attrs at Handle time.

Turns out this causes a minor violation in the slog contract. For example, for the following code:

```
func auditMiddleware(next sdk.Handler) sdk.Handler {
     return func(ctx context.Context, ic *sdk.InvocationContext) error {
         // top-level tenant_id, then a grouped "http" block — exactly the
         // pattern the dash0 guide and OTel semconv encourage for structured logs
         logger := slog.Default().
             With("tenant_id", tenantFromBaggage(ctx)).
             WithGroup("http").
             With(
                 "method", ic.TriggerMetadata["Method"],
                 "path",   ic.TriggerMetadata["Url"],
             )
 
         start := time.Now()
         err := next(ctx, ic)
         logger.InfoContext(ctx, "invocation finished",
             "duration_ms", time.Since(start).Milliseconds(),
         )
         return err
     }
 }
```
slog contract expects:
```
{
   "level": "INFO",
   "msg": "invocation finished",
   "tenant_id": "acme-corp",
   "http": {
     "method": "POST",
     "path": "/orders",
     "duration_ms": 142
   },
   "invocation_id": "...",
   "function_name": "hello",
   "trigger_type": "httpTrigger"
 }
```

what we return:
```
{
   "level": "INFO",
   "msg": "invocation finished",
   "http": {
     "tenant_id": "acme-corp",
     "method": "POST",
     "path": "/orders",
     "duration_ms": 142
   },
   "invocation_id": "...",
   "function_name": "hello",
   "trigger_type": "httpTrigger"
 }
```

ahmedmuhsin: Thanks, addressed in latest commits.

worker/handlers.go
ahmedmuhsin: Thanks, addressed in latest commits.

harshivcodes (approved):

worker/worker.go
```diff
@@ -1,34 +1,270 @@
...
+			//
+			// We deliberately don't try to construct a Failure
+			// InvocationResponse from the recovered value: the panicking
+			// path is far from the normal response-building code and
+			// reconstructing the InvocationId, RequestId, etc. here
+			// duplicates logic in handleInvocationRequest. The host
+			// already handles the missing-response case; surfacing the
+			// panic via the system logger is the most useful action.
+			defer func() {
+				if rec := recover(); rec != nil {
```
harshivcodes, line 240: If user code panics in any trigger that goes through the gRPC-body path (everything except httpTrigger routed through the HTTP-streaming proxy), the worker recovers the panic but never sends an InvocationResponse, so the host has to time out the invocation. 

Image

The HTTP-streaming path doesn't have this bug because its cross-goroutine rendezvous forced the author to convert panics into a pb.StatusResult for the channel handoff — a side effect that produces the right behavior for free.

Image

The panic will be surfaced via two distinct logs in app insights but not easily correlated:

Image

worker/log/doc.go
```diff
@@ -0,0 +1,68 @@
+// Package log implements the worker's logging pipeline: the wire
```
gavin-aguiar: Why do we need this file?

worker/log/observer_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"log/slog"
+	"sync"
+	"testing"
+	"time"
+
+	pb "github.com/azure/azure-functions-golang-worker/worker/proto"
+)
+
+// Each test in this file mutates the package-global observers
+// slice. Running them in parallel with anything else that registers an
```
gavin-aguiar, line 14: this is an observer test file but I don't see an observer.go.  Are we missing a file?

worker/handlers_middleware_test.go
```diff
@@ -0,0 +1,281 @@
+package worker
```
gavin-aguiar: Are we missing a handlers_middleware.go file?

gavin-aguiar: If its just testing what we have in handlers.go then these tests can go in handlers_test.go

worker/log/handlers.go
```diff
@@ -0,0 +1,370 @@
...
+// into the JSON properties bag.
+//
+// The fan-out to observers handles the case where the host has
+// WorkerOpenTelemetryEnabled=true and stops forwarding RpcLog User
+// records to its own telemetry pipeline; without an observer hooked up,
+// user logs would silently disappear in OTel mode. The otelfunc package
+// registers an observer that bridges to the OTel LoggerProvider so OTel
+// users get the right behavior automatically; users who don't import
+// otelfunc never link any OTel code into their binary.
+type userHandler struct {
```
gavin-aguiar, line 100: Will the userhandler be use for SDK and function app logs? If yes, shouldn't it be different?

worker/log/observer_test.go
ahmedmuhsin: Extracted the `Observer` type, the `observers` atomic.Pointer, and `RegisterObserver` out of `handlers.go` into a new `worker/log/observer.go` so the source/test pairing is explicit. Resolved in abc6358.

worker/handlers_middleware_test.go
ahmedmuhsin: The tests moved into `handlers_test.go` and the file was deleted (322ca0a).

worker/log/handlers.go
ahmedmuhsin: `userHandler` is only for user (function app) code logs, tagged as `RpcLog_User`. Worker-internal logs use a separate type, `systemHandler` (via `log.NewSystem`), tagged as `RpcLog_System`.

The two are deliberately different:

- `userHandler` (RpcLog_User): also fans the record out to registered `Observer`s so `otelfunc` can bridge user logs into the OTel `LoggerProvider`. When the host runs with `WorkerOpenTelemetryEnabled=true` it stops forwarding User records to its own telemetry pipeline, so without an observer they would silently disappear.
- `systemHandler` (RpcLog_System): pure RpcLog emission, no observer fan-out. Worker internals should not appear in user-facing OTel telemetry.

The SDK installs `userHandler` as the package-level default base via `sdk.SetDefaultBaseHandler`, so anything user code writes through `slog` flows out as User. The worker dispatcher uses `systemHandler` directly via `disp.SystemLogger()` for its own diagnostics.

harshivcodes (approved):

gavin-aguiar (approved):

PR #34, opened 2026-05-12
Standardize worker runtime to native and clean up worker config

harshivcodes:
Summary

Aligns all sample apps and documentation on FUNCTIONS_WORKER_RUNTIME = "native" and updates worker.config.json / docs to reflect the platform-neutral app executable name.

Changes

 - samples/*/local.settings.json: Set FUNCTIONS_WORKER_RUNTIME to "native" in all 9 sample apps. Previously 7 used "golang" and 2 already used "native" — now consistent 
across the board.
 - worker.config.json: Changed defaultExecutablePath from "app.exe" to "app".
 - TECHNICAL_SPEC.md:
  - Removed worker.config.json from the user app file structure (it's not required in the user's project).
  - Updated func init description to generate local.settings.json with "native" instead of "golang".
  - Simplified deployment artifact lists to reference app / proxy (dropped .exe variants).

Motivation

The golang worker runtime identifier was inconsistent across samples and no longer matches the platform convention of using native.
Standardizing avoids confusion for users copying sample configs.

Testing

Documentation and configuration only — no code changes. Verified via grep that no stale "golang" runtime references or app.exe defaults remain in the updated files

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,11 +19,11 @@ The `azure-functions-golang-worker` repository provides the SDK and worker imple
 ### Prerequisites
 
 * [Go 1.24+](https://go.dev/dl/)
-* Custom [Azure Functions Core Tools](https://www.npmjs.com/package/@gaaguiar/azure-functions-core-tools) with Go worker support
+* [Azure Functions Core Tools](https://www.npmjs.com/package/azure-functions-core-tools) v4.12.0-preview.1 or later (includes Go worker support)
 
 ```bash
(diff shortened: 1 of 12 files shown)


gavin-aguiar (approved):

ahmedmuhsin (changes requested):
Readme should be updated to not suggest that users download the experimental core tools versions from npm. That version does not recognize native.

ahmedmuhsin (approved):

PR #41, opened 2026-06-04
Add manual-trigger release pipeline for atomic module tags

ahmedmuhsin:
## Summary

Adds a manually-triggered release pipeline that creates atomic module tags (root + every submodule) from a single commit, plus the supporting changes needed to make Go module consumers happy and the CI pipeline pass SSCA.

## Why manual trigger

Releasing is a deliberate, separately-authorized action. Only people with permission to run this pipeline in ADO can publish tags, even though many people can merge a version bump. That keeps "can merge a PR" and "can cut a release" as distinct privileges: a merged change (including a malicious or accidental one) can never publish a public git tag on its own. Releases are rare, high-stakes, and effectively irreversible once consumed by the module proxy, so a human-in-the-loop with its own authorization boundary is the right posture.

## Release flow

1. Open a PR that bumps each submodule's `require github.com/azure/azure-functions-golang-worker` line in `go.mod` to the target version (e.g. `v0.7.0`) and merge it. Any merge style works (merge commit, squash, rebase); the pipeline reads file contents, not the commit message or branch name.
2. Wait for `golang-worker.official` to succeed on the merge commit.
3. Queue this pipeline. Under **Resources**, select that successful `golang-worker.official` run; its commit is the only commit this pipeline will tag (the build gate: you can't release a commit that didn't build). Leave the `rootTag` parameter empty to detect the version from `go.mod`, or set it to force a version.
4. The pipeline reads the pinned version from each submodule `go.mod` at that commit. If all submodules agree on a real semver version, it validates the `go.mod` files, then creates and pushes `vX.Y.Z` plus `middleware/otelfunc/vX.Y.Z`, `otelcollector/vX.Y.Z`, and `triggers/blob/vX.Y.Z`, all from the same commit. If the submodules disagree, the version is a pseudo-version, or the tag already exists at a different commit, the run **fails loudly**; the only no-op is when the tag already exists at the same commit (a safe idempotent re-run).
5. The pipeline then verifies the freshly-pushed tags resolve and build the way an external consumer sees them (see "Post-publish verification" below).
6. The pipeline prints `https://github.com/Azure/azure-functions-golang-worker/releases/new?tag=vX.Y.Z`; a human creates the GitHub Release in the UI with release notes.

## Version override

Setting the `rootTag` parameter (e.g. `v0.7.0`) forces that version instead of detecting it from `go.mod`. Use this to re-publish a failed run, or to release from a commit whose `go.mod` files weren't updated as part of the change. The selected official run's commit is still the only commit that gets tagged, and the `go.mod` validation still applies.

## Why detect the version from `go.mod`

A release PR is, by definition, the PR that bumps the submodule `go.mod` files to a new version. Deriving the version from those files means the merge style doesn't matter and there's no branch-naming convention to remember. The operation is idempotent: selecting a commit whose pinned version already has a tag (at the same commit) is a safe no-op.

## Post-publish verification

After pushing, the pipeline proves the tags work for a real consumer: in an isolated scratch module and module cache, it runs `go get` for the root and every submodule at the release version (`GOPROXY=direct`, `GOSUMDB=off` to hit the GitHub VCS directly and bypass any proxy cache), then `go build`s each submodule's packages standalone. This is the first point in the lifecycle where a non-workspace build is both possible and meaningful, because the released submodules now require a root version that actually exists. It fails loudly (with cleanup guidance) if resolution or the build fails.

## Auth

Tag pushes use the credentials persisted by `checkout: self` (`persistCredentials: true`). This relies on the project's GitHub service connection having write scope on the repo; no separate PAT is needed. The pipeline fails loudly with HTTP 403 if that assumption ever breaks.

## Other changes in this PR

- **Fixes a stale `genproto` pin in the root `go.mod`** (monolithic `google.golang.org/genproto` from Jan 2023, replaced by the split `genproto/googleapis/rpc` module). `go mod why` confirms the main module only needs the split module (via `grpc/status`, used by `worker/proto`). This removes the genproto half of the workspace ambiguity at the source. The `go.work` genproto replace below is still required for now because the submodules pin root `v0.6.0-preview`, whose published `go.mod` still drags in the old monolith and cannot be edited; it can be dropped once the submodules re-pin to a release built from the fixed root.
- Removes the committed `replace github.com/azure/azure-functions-golang-worker => ../...` directives from each submodule `go.mod`. Local dev and CI now rely on a gitignored `go.work`; published consumers pull a real root version from the proxy.
- Lifts `go work init` out of per-submodule build/test steps into a single setup step at the top of the build and unit-tests jobs.
- Switches three `tests/consumption/Dockerfile.flex-test*` files from `golang:1.24` (Docker Hub) to `mcr.microsoft.com/oss/go/microsoft/golang:1.24` to satisfy SSCA.
- Adds a README note encouraging consumers to pin to tagged releases rather than building against `main`.

## Validation

- End-to-end pipeline run with `rootTag=v0.0.0-pipelinetest1` succeeded: created and pushed all four annotated tags from the same commit.
- Post-publish verification mechanics validated locally against those `v0.0.0-pipelinetest1` tags: `go get` + `go build` of the root and all three submodules from origin succeed (no ambiguity, all build).
- After the genproto root fix, the root builds and `go vet`s both standalone (`GOWORK=off`) and in workspace mode, and all three submodules build in workspace mode.
- The `go.mod` extractor was unit-tested against the real single-line `require` form used in this repo, against a synthetic `require ( ... )` block form, and against negative cases (no root require, empty file).
- Local `go vet` + `go build` pass on root and all submodules with `go.work` and without `replace` directives in submodule `go.mod` files. Removing `go.work` reproduces the expected ambiguity error on `triggers/blob`, confirming workspace mode is genuinely required for local dev.


eng/ci/templates/jobs/build.yml
```diff
@@ -26,32 +26,43 @@ jobs:
...
+    # Initialize a go workspace so submodules resolve the root module locally
+    # instead of from the proxy. This is required because:
+    #   * submodule go.mod files reference an unreleased root version
+    #   * the currently published root (v0.6.0-preview) still contains packages
+    #     that have since been split into submodules, causing ambiguity
+    # The genproto replace pins a newer revision compatible with the OTel
+    # collector dependency tree.
+    - script: |
+        go work init . ./middleware/otelfunc ./otelcollector ./triggers/blob
+        go work edit -replace google.golang.org/genproto@v0.0.0-20230110181048-76db0878b65f=google.golang.org/genproto@v0.0.0-20250528174236-200df99c418a
```
harshivcodes, line 38: why are we editing this?

harshivcodes (approved):

eng/ci/templates/jobs/build.yml
ahmedmuhsin: This line works around the genproto module split. The OTel collector deps require the newer split modules (`genproto/googleapis/rpc`, `.../api`), while the root (through the `v0.6.0-preview` that the submodules pin) still pulls in the old monolithic `genproto`. In workspace mode both end up in one build, so the same import path exists in two modules and you get an "ambiguous import" error. The replace bumps the monolith to a post-split revision that no longer contains those packages, which removes the conflict.

I just fixed the root's own stale genproto pin in this PR (commit 10b8743), but this replace is still needed for now because `v0.6.0-preview`'s published go.mod can't be changed. Once the submodules re-pin to a release built from the fixed root, we can drop it.

harshivcodes (approved):

gavin-aguiar (approved):

PR #43, opened 2026-06-16
Add panic-safe goroutine helpers to the SDK

harshivcodes:
**Problem**

 The worker only recovers panics that happen synchronously in a handler. A panic in a customer-spawned  go func() { ... }()  runs on a different stack, escapes that  recover, and crashes the entire worker — failing every in-flight invocation with only an opaque "worker exited" error. This is the failure class from a recent  observation in a production app (nil-pointer panic in an EventHub handler's goroutine).

 **Changes**
 •  sdk/safego.go 
   •  Recover(ctx)  — deferrable guard for goroutines you launch yourself (WaitGroup, pool, named func). Logs and swallows the panic, keeping the worker alive.
   •  RecoverTo(ctx, &err)  — deferrable guard that captures the panic as an error via a pointer, so the invocation fails properly and the host can retry. Preserves any      pre-existing error.
 •  sdk/safego_test.go  — stdlib-only unit tests covering both helpers (7 test cases).
 •  worker/handlers_safego_test.go  — integration tests verifying the worker stays alive after a panicking goroutine.
 •  README.md  — new "Goroutine safety & panic recovery" section with  RecoverTo  and  Recover  usage examples.

 **Notes**

 • Panics are logged with full stack + invocation metadata; worker stays up.
 • Both helpers document that they must be the first  defer  (LIFO ordering ensures they run last and catch everything).
 • Stdlib-only, opt-in.

How this error propagates:
image

```
recovered panic in goroutine trigger_type=httpTrigger panic="runtime error: invalid memory address or nil pointer dereference" stack="goroutine 45 [running]:\nruntime/debug.Stack()\n\tC:/Program Files/Go/src/runtime/debug/stack.go:26 +0x5e\ngithub.com/azure/azure-functions-golang-worker/sdk.RecoverTo({0xafce10, 0x35b4d94ef3b0}, 0x35b4d9225f20)\n\tC:/Users/hakkaraj/Workspace/azure-functions-golang-worker/sdk/safego.go:92 +0x3e\npanic({0x9f4160?, 0xffc020?})\n\tC:/Program Files/Go/src/runtime/panic.go:860 +0x13a\nmain.RecoverToHandler.func1()\n\tC:/Users/hakkaraj/Workspace/azure-functions-golang-worker/samples/recoverTo/main.go:33 +0x63\ngolang.org/x/sync/errgroup.(*Group).Go.func1()\n\tC:/Users/hakkaraj/go/pkg/mod/golang.org/x/sync@v0.19.0/errgroup/errgroup.go:93 +0x50\ncreated by golang.org/x/sync/errgroup.(*Group).Go in goroutine 4\n\tC:/Users/hakkaraj/go/pkg/mod/golang.org/x/sync@v0.19.0/errgroup/errgroup.go:78 +0x95\n"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,33 +144,18 @@ The worker recovers panics that happen **inside your handler** and reports them 
 
 That safety net does **not** extend to goroutines you start yourself. In Go, an unrecovered panic in *any* goroutine terminates the entire process. In an Azure Functions worker this is especially dangerous: the worker hosts multiple concurrent invocations (potentially across different functions), so **one panicking goroutine crashes every in-flight request on the worker**, not just the function that spawned it. The host sees "worker exited" and the real stack never reaches your logs.
 
-### The pattern: `errgroup` + named-return recover
+### Propagate panics as errors: `sdk.RecoverTo`
 
-For work that matters (processing events, calling APIs), use [`errgroup`](https://pkg.go.dev/golang.org/x/sync/errgroup) with a named return so panics propagate as errors and the invocation fails properly — triggering retries instead of silent data loss:
+For work that matters (processing events, calling APIs), use [`sdk.RecoverTo`](sdk/safego.go) so panics propagate as errors and the invocation fails properly — triggering retries instead of silent data loss:
 
 ```go
-import (
-    "fmt"
-    "log/slog"
-    "runtime/debug"
-
-    "golang.org/x/sync/errgroup"
-)
+import "golang.org/x/sync/errgroup"
 
 func EventHubHandler(ctx context.Context, events []bindings.EventHubEvent) error {
     g, ctx := errgroup.WithContext(ctx)
     for _, e := range events {
         g.Go(func() (err error) {
-            defer func() {
-                if r := recover(); r != nil {
-                    // Logged with invocation_id, function_name, trigger_type
-                    // via the SDK's slog handler.
-                    slog.ErrorContext(ctx, "panic processing event",
-                        slog.Any("panic", r),
-                        slog.String("stack", string(debug.Stack())))
-                    err = fmt.Errorf("panic processing event: %v", r)
-                }
-            }()
+            defer sdk.RecoverTo(ctx, &err)
             return process(e)
         })
     }
@@ -178,7 +163,7 @@ func EventHubHandler(ctx context.Context, events []bindings.EventHubEvent) error
 }
 ```
 
-The `slog.ErrorContext(ctx, ...)` call is the Functions-specific piece: the SDK installs its log handler as the default, which pulls invocation metadata off `ctx` and attaches `invocation_id`, `function_name`, and `trigger_type` — so the panic shows up correlated in Application Insights.
+`RecoverTo` logs the full stack trace with invocation metadata (via slog + the SDK log handler) and sets `*errp` to a descriptive error. If `*errp` already holds a non-nil error, the original is preserved.
 
 ### Best-effort work: `sdk.Recover`
 
@@ -195,7 +180,7 @@ go func() {
 wg.Wait()
 ```
 
-> **Rule of thumb:** if a goroutine's failure should fail the invocation, use errgroup with a named-return recover. If the goroutine is truly best-effort, `sdk.Recover` keeps the worker alive with one line.
+> **Rule of thumb:** if a goroutine's failure should fail the invocation, use `sdk.RecoverTo`. If the goroutine is truly best-effort, `sdk.Recover` keeps the worker alive with one line.
 
 ---
 
--- a/README.md
+++ b/README.md
@@ -142,34 +142,60 @@ This is similar to the .NET worker extensions model (`Microsoft.Azure.Functions.
 
 The worker recovers panics that happen **inside your handler** and reports them to the host as a failed invocation with the full Go stack trace — so they show up, diagnosable, in Application Insights.
 
-That safety net does **not** extend to goroutines you start yourself. In Go, an unrecovered panic in *any* goroutine terminates the entire process. A handler that launches background work with a bare `go func() { ... }()` therefore puts the whole worker at risk: one nil-pointer dereference on that goroutine crashes the process, fails every other in-flight invocation with an opaque "worker exited" error, and the real stack never reaches the host or your App Insights.
+That safety net does **not** extend to goroutines you start yourself. In Go, an unrecovered panic in *any* goroutine terminates the entire process. In an Azure Functions worker this is especially dangerous: the worker hosts multiple concurrent invocations (potentially across different functions), so **one panicking goroutine crashes every in-flight request on the worker**, not just the function that spawned it. The host sees "worker exited" and the real stack never reaches your logs.
 
-Use [`sdk.Go`](sdk/safego.go) instead of `go` for fire-and-forget work. It runs your function in a panic-guarded goroutine: a panic is recovered, logged at error level with its stack (correlated to the spawning invocation via `slog`), and the worker stays alive.
+### The pattern: `errgroup` + named-return recover
+
+For work that matters (processing events, calling APIs), use [`errgroup`](https://pkg.go.dev/golang.org/x/sync/errgroup) with a named return so panics propagate as errors and the invocation fails properly — triggering retries instead of silent data loss:
 
 ```go
+import (
+    "fmt"
+    "log/slog"
+    "runtime/debug"
+
+    "golang.org/x/sync/errgroup"
+)
+
 func EventHubHandler(ctx context.Context, events []bindings.EventHubEvent) error {
+    g, ctx := errgroup.WithContext(ctx)
     for _, e := range events {
-        e := e
-        sdk.Go(ctx, func() {
-            // A panic in here is recovered and logged — it will NOT crash the worker.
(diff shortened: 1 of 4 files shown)


gavin-aguiar (approved):

ahmedmuhsin (changes requested):
The failure mode this goes after is real, and worth chasing. A panic in a customer goroutine taking down the whole worker, and every other in-flight invocation with it, is about as nasty as it gets, and pinning it to a clean repro is the hard part of the work. So I want to be clear that what follows isn't about the quality of the code, which is fine. It's about whether this belongs in the SDK as public API at all, and I landed on no. The reasoning is worth more than the verdict here, so let me walk through how I got there.

Short version, I don't think it should ship as public API. The helper makes a rarely-correct pattern (detached, swallow, continue) easy, the correct version is basically errgroup/conc plus one log line, and the only genuinely Functions-specific piece is that one line, which belongs in docs. The longer reasoning builds in order, so let me go step by step.

**1. The README example is actually an antipattern.**

```go
func EventHubHandler(ctx context.Context, events []bindings.EventHubEvent) error {
    for _, e := range events {
        e := e
        sdk.Go(ctx, func() {
            process(e)
        })
    }
    return nil
}
```

It spawns a goroutine per event and returns nil right away, without waiting for any of them. For EventHub, returning nil tells the host the batch succeeded, so the checkpoint moves past these events. But `process(e)` hasn't finished, and might not have even started. If it panics, `sdk.Go` recovers and logs it, except the event is already checkpointed and gone. A crash that should have triggered a retry quietly turns into lost data. And fire-and-forget work that outlives the invocation isn't guaranteed CPU, since the worker can be throttled or recycled between invocations, so the goroutine might not run at all.

So the flagship example for a safety helper actually makes things less safe. That's the symptom. The cause is the next two steps, and it goes deeper than the example.

**2. Adding `Wait` doesn't fix it.**

The obvious fix is to wait for the goroutines. But waiting by itself isn't enough. Here's the same recover-and-log pattern inside an errgroup that does wait.

```go
g, ctx := errgroup.WithContext(ctx)
for _, e := range events {
    g.Go(func() error {
        defer func() {
            if r := recover(); r != nil {
                slog.ErrorContext(ctx, "panic", slog.Any("panic", r)) // log only
            }
        }()
        return process(e)
    })
}
return g.Wait()
```

If `process(e)` panics, the defer recovers and logs, the function returns its zero value (nil error), errgroup sees nil, `Wait()` returns nil, and the invocation reports success. Same silent data loss, now with a Wait in front of it. The problem was never the missing Wait. It's that recovering and only logging swallows the failure, so the invocation reports success for work that panicked. That's true whether you wait or not.

I had to read this a couple of times to convince myself it was wrong, which is sort of the point. It looks correct.

**3. The fix is to propagate the panic, not swallow it.**

To make a recovered panic actually fail the invocation, you turn it into a returned error, using a named return so the deferred recover can set it.

```go
g, ctx := errgroup.WithContext(ctx)
for _, e := range events {
    g.Go(func() (err error) {
        defer func() {
            if r := recover(); r != nil {
                slog.ErrorContext(ctx, "panic processing event",
                    slog.Any("panic", r),
                    slog.String("stack", string(debug.Stack())))
                err = fmt.Errorf("panic processing event %q: %v", e.ID, r)
            }
        }()
        return process(e)
    })
}
return g.Wait() // non-nil on panic, so the invocation fails and the host retries
```

Now a panic becomes a non-nil error, `Wait()` returns it, the invocation fails, and the host retries the batch instead of checkpointing past it. (EventHub redelivers the whole batch on failure, so the handler needs to be idempotent, but that's true for any batch trigger.)

The part doing the real work is the named-return conversion, and a recover helper can't do that for you, because only the goroutine body knows how to describe its own failure. That's also why conc re-panics on `Wait()` instead of swallowing. Re-raising is how it forces the failure to surface. The libraries that wrap goroutines deliberately don't swallow.

**4. The correct version is errgroup plus a recover, and the valuable half isn't ours to build.**

Step 3 breaks into two parts. There's the lifecycle, which means spawning the goroutines, waiting for them, propagating the first failure, and cancelling the rest. And there's a recover to turn a panic into an error.

The lifecycle is the part that makes concurrent work safe, and it's exactly what errgroup and conc already give you. It's also the part `sdk.Go` doesn't provide at all. There's no Wait, no error propagation, no cancellation. What `sdk.Go` adds over a bare `go` is just the recover, and it implements it as swallow-and-continue, which is the one behavior both libraries deliberately avoid (errgroup doesn't recover at all; conc recovers and re-panics on `Wait()`).

So this isn't a leaner errgroup. It's the spawn-and-forget half with none of the lifecycle that makes it safe, plus the recover semantics you don't want. There's no gap for the SDK to fill here. The valuable part already exists upstream, and the part we'd be adding is the part people shouldn't reach for.

**5. The only Functions-specific part is one line.**

Look at what's actually unique to Functions in the correct version. Almost all of it is plain Go, like `go`, `defer`, `recover`, `debug.Stack`, named returns, and `fmt.Errorf`. Here's the only Functions-specific bit.

```go
slog.ErrorContext(ctx, "panic processing event", slog.Any("panic", r))
```

It's special only because the SDK installs its slog handler as the default, and that handler pulls the InvocationContext off `ctx` and attaches `invocation_id`, `function_name`, and `trigger_type`. So logging through `slog.ErrorContext` with the invocation's `ctx` is what makes the panic show up correlated in App Insights. Nothing else in the snippet depends on anything we ship.

And `Recover(ctx)` is basically that line plus a `recover()`.

```go
if r := recover(); r != nil {
    slog.ErrorContext(ctx, "...", panic, stack)
}
```

So the public API is wrapping a standard `defer recover()` and one log line behind three exported symbols (`Go`, `Recover`, `SetGoroutinePanicHandler`) plus a process-wide mutable handler that we then own forever.

**6. The convenience is aimed at the wrong thing.**

Most of the boilerplate the helper removes isn't accidental ceremony. It's the user stating real intent. Do I wait, what counts as failure, do I cancel the siblings. Those are exactly the decisions that produced the data-loss bug when they were hidden. A convenience wrapper is good when the thing it makes easy is the thing you should do. Here the thing it makes easy (detached, swallow, continue) is the thing you should almost never do. The only truly accidental boilerplate is that one telemetry line, and that's a doc snippet you drop into errgroup or conc, not an API.

There's a narrow exception, which is genuinely best-effort work where failure really doesn't matter (opportunistic cache warm, best-effort telemetry). Swallow-and-continue is fine there. But that's a small, deliberate case, and it doesn't justify a headline API that people reach for by default on work that does matter, like EventHub processing.

**7. We don't do this for the other workers, and there's a reason.**

Go is the outlier. An unrecovered panic in any goroutine takes down the whole process, and Go has no process-wide panic hook. The other workers don't need a helper like this because their runtimes already contain the blast radius of background failures.

| Worker | Background failure | Worker survives? |
|---|---|---|
| Go | panic in any goroutine | No, whole process dies |
| Python | exception in a thread or asyncio task | Yes, logged, that task dies |
| Java | uncaught exception in a thread | Yes, JVM runs the thread's handler and continues |
| .NET | unobserved Task exception | Yes for the Task path; raw Thread / threadpool still crashes |

I went through the invocation path in all three. Each catches synchronous handler errors and returns a Failure (same baseline as our `runUserInvocation`), and none of them registers a global uncaught-exception handler or ships a safe-thread / fire-and-forget helper. They lean on runtime defaults. To be fair, all of them still have a fatal tier they don't catch (Java `Error`, .NET fatal exceptions, Python `BaseException`, our non-recoverable panics), so none of this is blanket protection either.

So yes, we're not doing this for the other workers, and the reason is they don't have to. We'd be hand-building, as opt-in user API, the isolation those runtimes give for free.

**8. The one thing actually worth telling users.**

"A goroutine panic kills the process" is standard Go knowledge. The part that isn't standard, and is genuinely specific to us, is the multi-tenant angle. On this worker a panic doesn't just crash you, it fails the in-flight invocations of other functions sharing the same worker process. A Go dev reasonably assumes "I'll only crash myself, that's my problem," and doesn't realize the blast radius is shared. That's worth a clear paragraph in the docs. It's education, not API.

**What I'd suggest**

- Drop the public `Go` / `Recover` / `SetGoroutinePanicHandler` and replace it with a README section showing the pattern from step 3 (built on errgroup, with a note about conc for anyone who wants the panic re-raised on `Wait()`), and call out the one Functions-specific line. That gives users real join and error propagation, fixes the data loss, and adds nothing permanent to the SDK.
- If we want a single place to recover panics in our own goroutines later (OTel exception events, metrics), keep it as an unexported helper. Internal detail, not public API.
- Either way, the EventHub example needs fixing, since it's the thing people will copy. The step 3 version is the one I'd ship.

I could be missing an internal use case for keeping `Recover` exported, so happy to be talked out of it. But as it stands, I think this is mostly a problem the ecosystem already solved correctly, plus one line of telemetry glue that belongs in the docs instead of the API.

ahmedmuhsin (approved):
Ordering fix looks right. `RecoverTo` now leads with errgroup and the hand-rolled WaitGroup example registers `wg.Done` before `RecoverTo` so the error is set before `Wait` unblocks. That's the correct LIFO order and it matches the integration test.

Thanks for working through all of it. LGTM.

PR #44, opened 2026-06-18
feat(durable): add Durable Functions support to the Go worker

### LINKED ISSUE #47: [Go Worker] Add Durable Functions Support
Track Durable Functions support in the Go worker.

Planning epic: Azure/azure-functions-bucees-planning#1131 (this issue is the worker-repo tracking issue for that effort).

## Approach
Execution uses **trigger-replay (Model 1)**, matching the dotnet-isolated worker: the host's DurableTask extension owns state and dispatch; the Go worker is a stateless replay engine. A single `app.Use(durabletask.Middleware(...))` registers orchestrators/activities and replays orchestration invocations. Management (start / status / raise-event / terminate) uses the durable gRPC client, delivered to starters via the `durableClient` binding. Tracing is native through the otelfunc middleware.

Validated end-to-end on a real Functions host (Azure Functions Core Tools 4.12.0, native Go worker) and against live Application Insights. Delivery is split into **three phases** so the worker can ship independently of upstream durabletask-go release cadence.

## Phases
1. **Connect existing support** — Durable works on the Go worker: orchestration replay, activities, fan-out/fan-in, human-in-the-loop, management binding, and worker-side OpenTelemetry (per-invocation spans + per-turn diagnostics). orchestration → activity → worker cascade; starter → orchestration bridged by instance-id correlation until Phase 2.
2. **Distributed-tracing context cascade** — propagate caller trace context on `ScheduleNewOrchestration` (parity with .NET) for a single starter → orchestration → activity trace. A durabletask-go change plus a one-line worker dependency bump.
3. **Durable Entities** — `entityTrigger` support (durabletask-go entity engine + worker binding + extension hookup).

## PR map
| Phase | Repo | PR |
|---|---|---|
| 1 | azure-functions-golang-worker | #44 |
| 1 | azure-functions-durable-extension | Azure/azure-functions-durable-extension#3457 |
| 2 | durabletask-go | microsoft/durabletask-go#134 |
| 2 | azure-functions-golang-worker | (dependency bump — follow-up after durabletask-go release) |
| 3 | durabletask-go + worker + extension | (TBD) |

ahmedmuhsin:
## Summary

Adds Durable Functions to the Go worker behind a single `app.Use(durabletask.Middleware(...))`. Orchestrators and activities run through the normal Functions trigger pipeline, the host's DurableTask extension stays the backend (state, dispatch, timers, queues), and the worker is a stateless replay engine.

This PR doubles as the design doc for durable support in the Go worker, so it spends most of its length on the architecture and the rationale rather than the diff.

It is Phase 1 of three (planning section at the bottom). It is deliberately fork free. It builds against `durabletask-go v0.6.0` with no `replace` directive, so it can merge on its own without waiting on any upstream release.

The companion host change, Azure/azure-functions-durable-extension#3457, has merged and shipped in DurableTask extension 3.15.0. It teaches the host to recognize the native worker runtime and select the gRPC durable protocol, which durable bindings need in order to dispatch.

The sample pins the experimental extension bundle, currently the only bundle whose CDN index resolves to 3.15.0 or later. Both bundles are served from the same default CDN, so no source URI override is needed. Once a standard bundle catches up the sample can move back to it.

## Usage

The whole feature is one `app.Use`. Orchestrators and activities are registered on the middleware, and an HTTP starter reaches the management client through the `durableClient` binding.

```go
func main() {
    app := sdk.FunctionApp()

    // otelfunc first so it wraps every invocation with a span.
    app.Use(otelfunc.Middleware())

    // Register the orchestrators and activities, then hand the middleware to
    // the app. Registering after app.Use panics, since that is when the app
    // collects them.
    durable := durabletask.Middleware()
    durable.Orchestrator("HelloCities", HelloCities)
    durable.Activity("SayHello", SayHello)
    app.Use(durable)

    // ClientInput() adds the durableClient binding so the starter can reach the client.
    app.HTTP("StartHelloCities", StartHelloCities,
        sdk.WithMethods("post"), sdk.WithRoute("hello"), durabletask.ClientInput())

    worker.Start(app)
}

// Orchestrator. Deterministic, drives the workflow by scheduling activities.
func HelloCities(ctx *task.OrchestrationContext) (any, error) {
    var greetings []string
    for _, city := range []string{"Tokyo", "Seattle", "London"} {
        var greeting string
        if err := ctx.CallActivity("SayHello", task.WithActivityInput(city)).Await(&greeting); err != nil {
            return nil, err
        }
        greetings = append(greetings, greeting)
    }
    return greetings, nil
}

// Activity. An ordinary function, non-deterministic code is fine here.
func SayHello(ctx context.Context, city string) (string, error) {
    return fmt.Sprintf("Hello, %s!", city), nil
}

// HTTP starter. Schedules an instance via the durable client from the request context.
func StartHelloCities(w http.ResponseWriter, r *http.Request) {
    client, _ := durabletask.ClientFromContext(r.Context())
    id, _ := client.ScheduleNewOrchestration(r.Context(), "HelloCities", nil)
    // 202 plus the host management URLs, the same reply the other workers send.
    _ = client.WriteCheckStatusResponse(w, r, id)
}
```

## Execution model, and why it is trigger-replay

There are two ways an out-of-process worker can run durable code.

1. **Trigger-replay (Model 1).** The host owns the backend and dispatches each orchestrator turn as an ordinary function invocation. The trigger payload is the orchestration history, the worker replays the orchestrator and returns the resulting actions. This is what the dotnet-isolated worker does via `GrpcOrchestrationRunner.LoadAndRun`.
2. **Work-item stream (Model 2).** The worker opens its own gRPC stream to a backend sidecar and pulls work items itself. This is the standalone durabletask-go / DTS topology.

I went with Model 1 because that is what the Functions host actually drives. The host's `TaskHubGrpcServer` implements only the client and management RPCs (start, raise event, terminate, query, purge). It has no `GetWorkItems` stream, and its `OutOfProcMiddleware` invokes the orchestrator through the normal FunctionRpc trigger pipeline, reading the worker's return value as the base64 `OrchestratorResponse`. Model 2 would only work against a standalone backend, not the Functions host. The management client is still gRPC, so the final shape is trigger-replay for execution plus a gRPC client for management.

## How orchestration replay works

The orchestrator invocation arrives as an `orchestrationTrigger` whose input is a base64 `OrchestratorRequest` (past events plus new events). The middleware short circuits it in `Wrap` and hands it to `loadAndRun`, which does three things.

1. Decode the request. `durabletask-go` keeps its protobuf types in `internal/protos`, so the request envelope is decoded field by field with `protowire` (instance id, past events, new events). Each history event is handed to the public `backend.UnmarshalHistoryEvent`.
2. Replay. `task.NewTaskExecutor(registry).ExecuteOrchestrator(...)` runs the registered orchestrator over the history and produces an `OrchestratorResponse` (the actions to take).
3. Encode the response. The response is marshaled and set as the invocation's return value, which the host's DurableTask extension applies.

Orchestrator code uses the durabletask-go programming model (`task.OrchestrationContext`, `CallActivity`, `CreateTimer`, `WaitForSingleEvent`, `SetCustomStatus`, `ContinueAsNew`). It must stay deterministic, the same rule as every other language.

## How activities work

Activities are ordinary Functions. Input in, result out, no replay. The middleware lets the activity invocation fall through `Wrap` to the normal handler path. The one wrinkle is input decoding. The host delivers the raw scheduled input, which is JSON encoded, so the middleware JSON-decodes a scalar string into the activity's typed parameter rather than handing the handler a quoted string. That fix is its own commit in the history.

## Management client and the durableClient binding

Starter and management functions (start an orchestration, query status, raise an event) need the durable management client. The host hands the worker its local gRPC endpoint through a `durableClient` input binding. A starter opts in with `durabletask.ClientInput()` on its registration, the middleware reads the delivered `rpcBaseUrl`, dials the gRPC client once per endpoint, caches it, and attaches it to the request context. Handlers reach it with `durabletask.ClientFromContext(r.Context())`.

This is the same path the dotnet-isolated worker uses. `BindingHelper.DurableOrchestrationClientToString` serializes the local RPC address into the binding value, and the worker parses it back out.

The binding also carries the durable webhook base URL and the query string management calls require, so `Client.WriteCheckStatusResponse` can return the canonical starter reply: HTTP 202, the instance management URLs, a `Location` header and a `Retry-After` hint. That matches the `CreateCheckStatusResponse` helper every other language worker ships. `Client.ManagementURLs` exposes the same payload for callers that want to shape their own response.

## Worker core seams

Durable needs a few capabilities that the core worker did not expose yet. I kept them general rather than durable-specific.

- **Return-value encoding for non-HTTP triggers.** The dispatcher now captures the first non-error return from a handler and encodes it into `InvocationResponse.ReturnValue`. Every other worker already does this for all trigger types, the Go worker only did it for the HTTP `$return`. The orchestrator response rides this path.
- **Raw trigger input bytes.** `MiddlewareContext.SetInputBytes` / `InputBytes` exposes the untouched trigger payload, which is how the base64 `OrchestratorRequest` reaches the replay engine.
- **Non-trigger binding inputs.** `MiddlewareContext.SetBindingInput` / `BindingInput` surfaces additional input bindings (the `durableClient` endpoint) onto the context. `populateBindingInputs` wires this for both the gRPC body path and the HTTP-streaming proxy path.
- **`FunctionProvider`.** The host only dispatches functions it learned about during worker indexing, so each orchestrator and activity has to be registered on the `App` with its `orchestrationTrigger` or `activityTrigger` binding, not just held inside the middleware. `FunctionProvider` is the seam that lets a middleware hand its own function registrations to the `App` at `app.Use` time. So `WithOrchestrator` / `WithActivity` populate two registries, the durabletask-go `task.TaskRegistry` that the replay engine executes against, and (through `FunctionProvider`) the `App`'s function list that becomes the host metadata. Without this seam you would have to declare every orchestrator and activity twice, once for replay and once again with an `app` registration call so the host knows the trigger exists. The registered body of an orchestrator is a no-op placeholder, since `Wrap` intercepts the orchestration invocation and runs the replay itself, the placeholder only exists so the host emits metadata and dispatches the `orchestrationTrigger`. Because `app.Use` collects those registrations at the moment it is called, registering afterwards now panics with an actionable message rather than leaving an orchestration the host never learns about.

## Tracing

Tracing is native through the existing `otelfunc` middleware, the durable middleware has no OpenTelemetry import of its own.

- Every invocation (starter, orchestrator turn, activity) gets a worker span from `otelfunc`.
- On the orchestrator replay span I attach per-turn diagnostics so replay is legible in the backend: `durabletask.is_replay`, `durabletask.history_event_count`, `durabletask.new_events_count`, `durabletask.action_count`, `durabletask.task.instance_id`. The names follow durabletask-go's own `durabletask.*` convention. It is self gating, if an app does not register `otelfunc` the active span is non-recording and the annotation is a no-op, so apps that opt out pay nothing.

What is in this PR covers the orchestration to activity to worker cascade in one trace. The starter to orchestration link is Phase 2, it needs a durabletask-go change to propagate the caller trace context on schedule. Until that lands you can still correlate the starter and the orchestration by `durabletask.task.instance_id`, which is on the worker orchestration span today.

## Sample

`samples/durableFunctions` shows the whole surface.

- **HelloCities**, a sequential orchestration calling `SayHello` per city.
- **ProcessExpense**, fan-out / fan-in across three validation activities, custom-status progress, and a human-in-the-loop approval gated by a durable timeout.
- The management endpoints (start, status, raise event) and the OpenTelemetry wiring.

## Testing

- Unit and emulator-backed tests for the middleware, the replay decode/encode round trip (asserted byte-for-byte against the engine's own output), the `durableClient` binding delivery, the registration ordering guard, the check-status response, and the turn-span annotation.
- `TestDurableOrchestrations` in `tests/integration` drives the sample through a real Functions host (official Core Tools 4.12.0, native Go worker) and covers a sequential orchestration, fan-out / fan-in with automatic approval, human approval and rejection through an external event, rejection on failed validation, and an unknown instance lookup. It is wired into PR validation alongside the other integration scenarios.
- Every assertion goes through the sample's own anonymous HTTP endpoints rather than the host management API, so the same suite runs unchanged against a deployed app by setting `DURABLE_E2E_BASE_URL`. All six scenarios pass against a Flex Consumption app in Azure on the released extension, with no patched assemblies.
- Per-turn diagnostics confirmed live in Application Insights.

## Phasing

Durable is landing in three phases so the worker is not blocked on upstream release cadence.

1. **Phase 1, this PR plus the host change (#3457).** Durable works end to end. Orchestration to activity to worker tracing. Fork free.
2. **Phase 2, context cascade.** Propagate the caller trace context on `ScheduleNewOrchestration` for a single starter to orchestration to activity trace. That is a durabletask-go change (microsoft/durabletask-go#134) plus a one-line dependency bump here once it releases. No other worker change, the span and the context are already in place, they are just dropped by the v0.6.0 gRPC client today.
3. **Phase 3, entities.** `entityTrigger` support, which needs entity primitives in durabletask-go first.

## Related PRs

- Host runtime recognition (required companion): Azure/azure-functions-durable-extension#3457
- Phase 2 trace propagation: microsoft/durabletask-go#134


vrdmr:
Tracking issue: #47 — this PR is **Phase 1** of Durable Functions support in the Go worker. Planning epic: Azure/azure-functions-bucees-planning#1131.

harshivcodes (approved):
Left a suggestion - and filed an issue for my open concern.

middleware/durabletask/doc.go
```diff
@@ -0,0 +1,114 @@
+// Package durabletask provides Durable Functions support for the
```
gavin-aguiar: Can we make this a README.md ?

gavin-aguiar:
Few comments: 
1. High – Extended sessions can duplicate orchestration actions
 middleware/durabletask/runner.go:17-24,48-67,145  ignores request  properties  and always replays, even when the host omits history for an extended session. .NET returns  requiresHistory ; Go may replay empty history and reschedule completed activities. Parse this flag and request history, or explicitly reject extended sessions.
2. Medium – Incorrect management URLs behind proxies/custom domains
 middleware/durabletask/checkstatus.go:92-98  always prefers the binding URL over the HTTP request. .NET isolated and JavaScript prefer the caller-visible request origin, so Go can return unusable internal/default-host URLs behind APIM or custom domains.
3. Medium –  ProvidedFunctions  violates its interface contract
 sdk/middleware.go:159-160  says it is side-effect-free, but  middleware/durabletask/durabletask.go:277-281  closes registration. Innocent introspection makes subsequent registrations panic with the incorrect claim that  app.Use  was called. Move the close operation into  App.Use  or change the contract.
4. Medium – Payload-copy regression affects every function
 worker/handlers.go:300-303,493-521  and  worker/http_proxy.go:314-317  eagerly copy string/JSON trigger payloads even without middleware. Large Event Hub, Service Bus, or queue batches now consume significant additional memory. Populate lazily or only when middleware requires raw inputs.
5. Medium – Client API removes standard orchestration options
 middleware/durabletask/client.go:97-105  exposes only  (name, input) , hiding durabletask-go’s explicit instance ID, scheduled start, and ID-reuse options. Other language SDKs expose these because they are required for singleton/idempotent orchestration patterns. Functional options would also be more idiomatic Go.
6. Comments/telemetry do not match behavior
•  durabletask.go:98  documents nonexistent  WaitForExternalEvent ; v0.6.0 provides  WaitForSingleEvent .
•  runner.go:93  sets  durabletask.is_replay  whenever any past history exists, unlike the SDK’s point-in-time  IsReplaying ; almost every later turn is therefore mislabeled.
•  checkstatus.go:14-30  calls the payload standard but omits  rewindPostUri , which other workers include.

ahmedmuhsin:
@gavin-aguiar Thanks, useful pass. Four were right and are fixed. One I'd push back on, one I fixed differently.

**1. Extended sessions.** Not reachable. `DurableTaskOptions.Validate` throws at host startup when `extendedSessionsEnabled` is true and `FUNCTIONS_WORKER_RUNTIME` isn't `dotnet` or `dotnet-isolated`, and the Flex image bakes in `native`. Confirmed locally and on Flex, where `/admin/host/status` comes back `"state": "Error"` with that message. The check is in every 3.x release and every bundle a Go app can load carries a 3.x extension.

Added a guard anyway rather than implementing the protocol. A request always carries `ExecutionStarted` somewhere, so empty `pastEvents` with none in `newEvents` means the host withheld history, and we now fail loudly instead of re-running completed work. Checked that invariant against 16 real turns including ContinueAsNew and sub-orchestrations.

**2. Management URLs.** Real bug, opposite fix. The host forwards requests to the worker over loopback, so `r.Host` is the worker's own listener. On the deployed app it was `127.0.0.1:36833` while `X-Forwarded-Host` had the real name, and `r.TLS` was nil despite an HTTPS caller. Preferring the request would have made it worse. It now prefers the forwarded headers like the isolated worker does, falling back to the binding so a missing header can't produce a loopback URL. That fixes the custom domain case.

**3. `ProvidedFunctions`.** Agreed, and the panic message was misleading too. Added `sdk.RegistrationSealer`, which `App.Use` calls once it has taken the registrations.

**4. Payload copy.** Agreed, though narrower than described. HTTP triggers and batches cost nothing. The real cost was single string and JSON triggers, plus durable itself doing `string(mc.InputBytes())`, which was two copies of the history per turn. Text is carried as a string now and converted only if something asks for bytes, so durable is zero copies and matches the isolated worker.

**5. Client options.** Agreed, and went further. `Client` embeds `TaskHubGrpcClient`, so everything upstream offers is reachable including anything added later. The package keeps only what upstream can't know about, which is connection ownership and the Durable HTTP contract.

**6.** `WaitForSingleEvent` fixed. `rewindPostUri` added, it's preview rather than deprecated and every other worker emits it. On `is_replay` I removed it instead of aligning it. `IsReplaying` flips partway through a turn, so no single value is right for a span covering the whole turn, and neither durabletask-go nor the .NET SDK emits it.

Six commits, each building on its own. All modules green, and the e2e suite passes locally and against both deployed apps.

ahmedmuhsin:
Unrelated find while verifying this, worth flagging on its own.

`ctx.CallSubOrchestrator` without an explicit instance ID hung forever. The child was never created, the parent stayed Running, and nothing was logged anywhere.

durabletask-go leaves the ID to the backend and fills it in there:

```go
// backend/runtimestate.go
if createSO.InstanceId == "" {
    createSO.InstanceId = fmt.Sprintf("%s:%04x", s.instanceID, action.Id)
}
```

Under Functions the backend is the host's DurableTask extension, which does neither and forwards `InstanceId=""`, so the message goes nowhere. The .NET SDK handles this in the worker, where `CallSubOrchestratorAsync` falls back to a generated ID.

Fixed the same way, using durabletask-go's own format so an orchestrator produces identical child instance IDs under Functions and standalone.

middleware/durabletask/doc.go
ahmedmuhsin: Did both. `doc.go` is what pkg.go.dev renders, so a plain move would have left the package page empty. It now keeps the overview, the wiring example and the contracts, and there's a new README for the operational side (bundle requirement, Storage vs DTS, running locally, known gaps).

gavin-aguiar (approved):

PR #45, opened 2026-06-22
Add Azure Storage Queue Trigger Support

harshivcodes:
**Summary**

 Adds a  queueTrigger  binding so functions can be triggered by Azure Storage Queue messages, plus a converter fix ensuring queue/Service Bus messages with JSON bodies  populate correctly.

 **Changes**

 • Queue trigger SDK: New  QueueStorageTrigger ,  QueueBinding , and  QueueMessage  types;  app.Queue(...)  registration;  QueueHandler  handler type.
 • Options:  WithQueueName  and  WithConnection  now also apply to Storage Queue bindings.
 • Converter fix: Structs with an  azfuncdata -tagged field now skip the whole-struct JSON fast-path, so a JSON-object queue body is routed to  Body  (with metadata    fields populated separately) instead of silently producing a zero-valued message.
 • Decode robustness:  decodeProto  handles host extensions delivering values via different  TypedData  variants (e.g. queue  DequeueCount  as JSON number, time fields as    JSON strings, JSON bodies into  []byte / json.RawMessage ).
 • Telemetry: otel middleware classifies  queueTrigger  as  pubsub  and  sqlTrigger  as  datasource .

 **Tests**

 • New converter unit tests, queue binding tests, app registration tests, and a queue storage integration test.
 • Sample app under  samples/queueTrigger/ .

### DIFF
--- a/samples/queueTrigger/host.json
+++ b/samples/queueTrigger/host.json
@@ -1,0 +1,7 @@
+{
+  "version": "2.0",
+  "extensionBundle": {
+    "id": "Microsoft.Azure.Functions.ExtensionBundle",
+    "version": "[4.*, 5.0.0)"
+  }
+}
--- a/samples/queueTrigger/host.json
+++ b/samples/queueTrigger/host.json
@@ -1,5 +1,10 @@
 {
   "version": "2.0",
+  "extensions": {
+    "queues": {
+      "messageEncoding": "none"
+    }
+  },
   "extensionBundle": {
     "id": "Microsoft.Azure.Functions.ExtensionBundle",
     "version": "[4.*, 5.0.0)"
--- a/samples/queueTrigger/host.json
+++ b/samples/queueTrigger/host.json
@@ -1,10 +1,5 @@
 {
   "version": "2.0",
-  "extensions": {
-    "queues": {
-      "messageEncoding": "none"
-    }
-  },
   "extensionBundle": {
     "id": "Microsoft.Azure.Functions.ExtensionBundle",
     "version": "[4.*, 5.0.0)"
--- a/samples/queueTrigger/local.settings.json
+++ b/samples/queueTrigger/local.settings.json
@@ -1,0 +1,7 @@
+{
+  "IsEncrypted": false,
+  "Values": {
+    "AzureWebJobsStorage": "UseDevelopmentStorage=true",
+    "FUNCTIONS_WORKER_RUNTIME": "native"
+  }
+}
--- a/samples/queueTrigger/local.settings.json
+++ b/samples/queueTrigger/local.settings.json
@@ -2,6 +2,7 @@
   "IsEncrypted": false,
   "Values": {
     "AzureWebJobsStorage": "UseDevelopmentStorage=true",
-    "FUNCTIONS_WORKER_RUNTIME": "native"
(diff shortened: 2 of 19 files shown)


samples/queueTrigger/host.json
```diff
@@ -0,0 +1,12 @@
+{
+  "version": "2.0",
+  "extensions": {
```
gavin-aguiar: We can remove this since its not used/needed.

samples/queueTrigger/local.settings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "IsEncrypted": false,
+  "Values": {
+    "AzureWebJobsStorage": "UseDevelopmentStorage=true",
+    "FUNCTIONS_WORKER_RUNTIME": "native",
+    "FUNCTIONS_CLI_NATIVE_LANGUAGE": "go"
```
gavin-aguiar, line 6: What is this settings used for?

worker/converter_test.go
```diff
@@ -271,6 +395,43 @@ func TestConvertToTypeValue_TimerInfo_StringJSON(t *testing.T) {
 
 // --- encodeHTTPResponse tests ---
 
+// TestConvertToTypeValue_AzFuncData_JSONBody is a regression test for the
+// case where a queue (or service bus) message has a JSON object/array as
+// its body. Without the azfuncdata-aware guard in convertToTypeValue, the
+// whole-struct JSON fast-path would silently succeed against the body's
+// JSON (matching no field tags) and return an entirely empty struct —
+// the per-field metadata + azfuncdata fallback would never run.
+func TestConvertToTypeValue_AzFuncData_JSONBody(t *testing.T) {
```
ahmedmuhsin, line 404: This inline `queueMsg` test covers the queue case, but the PR also changes how `ServiceBusMessage` decodes since it carries the same `azfuncdata` tag, and that path has no direct coverage right now. Can we add a sibling test that runs a real `bindings.ServiceBusMessage` through `convertToTypeValue` with a JSON object body and a couple of metadata fields? That's the half most likely to regress silently later.

samples/queueTrigger/local.settings.json
harshivcodes: Its being used by core tools - it was added last minute. func init automatically adds it to the local.settings.json file.

samples/queueTrigger/host.json
harshivcodes: ack - removed it.

ahmedmuhsin (approved):

tests/integration/go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.21.1 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.12.0 // indirect
+	github.com/Azure/azure-sdk-for-go/sdk/storage/azqueue v1.0.1 // indirect
 	github.com/Azure/go-amqp v1.4.0 // indirect
```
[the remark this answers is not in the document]
gavin-aguiar: This comment is valid right?

gavin-aguiar (approved):

PR #48, opened 2026-07-06
feat(sdk/cosmos): add change-feed trigger configuration options

harshivcodes:
**Summary**

Adds full CosmosDB change-feed trigger configuration to the SDK, including the ability to auto-provision the lease container on first run.

**Changes**

• New option  WithCreateLeaseContainerIfNotExists  — opts the CosmosDB extension into creating the lease container automatically instead of requiring it to be provisioned out of band.
• 14 additional Cosmos trigger options covering the rest of the change-feed knobs exposed by the Functions host:
• Lease topology:  WithLeaseContainer ,  WithLeaseDatabase ,  WithLeaseConnection ,  WithLeaseContainerPrefix ,  WithLeaseContainerThroughput 
• Timing (all  time.Duration , truncated to ms):  WithFeedPollDelay ,  WithLeaseRenewInterval ,  WithLeaseAcquireInterval ,  WithLeaseExpirationInterval 
• Feed behavior:  WithMaxItemsPerInvocation ,  WithStartFromBeginning ,  WithStartFromTime ,  WithPreferredLocations ,  WithChangeFeedMode  (with typed  CosmosChangeFeedMode  enum)
• Sample — the existing  samples/cosmosDBTrigger  README documents the auto-lease option and links to the full option set

Tests

•  TestCosmosDB_AllOptions  — every option round-trips into the binding
•  TestCosmosDB_DurationOptions_TruncateToMilliseconds  — sub-ms truncation behavior
•  TestCosmosOptions_NonCosmosTriggerIsNoop  — Cosmos options are safe no-ops on other triggers
•  TestCosmosDBBinding_JSON_{Omits,Emits}OptionalFieldsWhenZero/Set  — wire-format contract with the host


gavin-aguiar (approved):

sdk/options.go
```diff
@@ -62,6 +67,138 @@ func WithContainer(containerName string) Option {
...
+}
+
+// WithChangeFeedMode selects the change-feed mode the trigger reads from.
+// Defaults to [bindings.CosmosChangeFeedModeLatestVersion]. Use
+// [bindings.CosmosChangeFeedModeAllVersionsAndDeletes] to receive every
+// intermediate mutation and delete event (requires the account feature to be
+// enabled).
+func WithChangeFeedMode(mode bindings.CosmosChangeFeedMode) Option {
+	return withCosmos(func(c *bindings.CosmosDBBinding) { c.ChangeFeedMode = mode })
+}
```
ahmedmuhsin: Let's pull AllVersionsAndDeletes out of this PR. That means this option, the CosmosChangeFeedMode enum, and the ChangeFeedMode field.

In that mode the host changes the payload shape. CosmosDBTriggerAttributeBindingProviderGenerator resolves the out-of-proc document type to ChangeFeedItem<JObject>, so each item arrives as {"current":...,"previous":...,"metadata":...} with current null on deletes. Our handler is fixed to []bindings.CosmosDocument, which only carries flat fields like id and data, so nothing matches. Every document decodes empty and deletes lose their identity.

I checked this against the real decode path, decodeProto JSON fallback into []CosmosDocument. LatestVersion gives ID[0]="doc-1" ID[1]="doc-2". AllVersionsAndDeletes gives ID[0]="" ID[1]="" Data[0]="". LatestVersion stays fine.

Let's ship the other 14 knobs and add this one back once we have a ChangeFeedItem shaped handler.

sdk/options.go
```diff
@@ -62,6 +67,138 @@ func WithContainer(containerName string) Option {
 	}
 }
 
+// withCosmos applies f to the trigger's CosmosDBBinding when one is present.
+// Options for other trigger types are silently no-ops on non-Cosmos triggers.
+func withCosmos(f func(*bindings.CosmosDBBinding)) Option {
```
ahmedmuhsin, line 54: Good helper. WithDatabase and WithContainer just above still inline the triggerBinding nil check. Let's move them onto withCosmos too so every Cosmos option goes through the same path.

sdk/options.go
```diff
@@ -62,6 +67,138 @@ func WithContainer(containerName string) Option {
...
+
+// WithPreferredLocations sets preferred regions for a geo-replicated CosmosDB
+// account. Values are joined with commas as required by the binding format,
+// e.g. WithPreferredLocations("East US", "North Europe").
+//
+// Empty strings are not filtered: WithPreferredLocations("East US", "")
+// serializes as "East US,", which the host may reject. Pass only non-empty
+// region names.
+func WithPreferredLocations(locations ...string) Option {
+	return withCosmos(func(c *bindings.CosmosDBBinding) { c.PreferredLocations = strings.Join(locations, ",") })
```
ahmedmuhsin: Rather than document the empty string footgun we can just remove it. Filter empties before the Join so ("East US", "") emits "East US" instead of "East US,", which the host may reject.

ahmedmuhsin (commented):
Solid PR and locking the wire format down with those JSON tests is the right call. One blocker on AllVersionsAndDeletes plus two small nits. Rest looks good to me.

ahmedmuhsin (approved):
Thanks for the fixes. Dropping AllVersionsAndDeletes, routing WithDatabase and WithContainer through withCosmos, and filtering empty locations all look good. Approving.

PR #49, opened 2026-07-07
ci: fix CodeQL Go analysis producing empty databases 

harshivcodes:
Description

CodeQL currently reports success on every build but ships an empty Go database ( No source code was built for go , 22-byte source archive). Root cause: the auto-injected CodeQL Init runs before Go is installed and before  go work init  creates the multi-module workspace, so the extractor never hooks the compiler [sample Pipeline run](https://dev.azure.com/azfunc/public/_build/results?buildId=287399&view=logs&j=6b58850f-3858-5a05-33e2-5e41cbf03c4e&t=c934e8dd-283f-54d0-1dc6-bfbb6611c687).

Changes:

• Move  GoTool@0  +  go work init  into  templateContext.preSteps  so they run before CodeQL Init.
• New  eng/ci/scripts/codeql-build.sh  does one traced build of the root + all submodules + samples, replacing the per-submodule build steps.
• New  CodeQL.yml  classifies  *_test.go ,  worker/proto/** ,  samples/** , and  tests/**  to reduce noise.
• Disable CodeQL on the Unit Tests job (redundant with Build job,  go test  doesn't produce a usable trace).
• Add a  postSteps  guard that fails the Build job if the resulting CodeQL Go DB is essentially empty, to prevent future silent regressions.
• Add  .gitattributes  to keep  *.sh  LF on Linux agents.

```
Before:

┌─ Build Go Worker job ─────────────────────────────┐
│                                                   │
│  🛡 CodeQL Initialize   ◄── runs first, no Go     │
│  Install Go 1.25                                   │
│  Initialize go workspace                           │
│  go build (per submodule)  ◄── tracer misses it   │
│  🛡 CodeQL Finalize  →  empty DB, warning only    │
│                                                   │
└───────────────────────────────────────────────────┘
```
```
Now:
┌─ Build Go Worker job ─────────────────────────────┐
│                                                   │
│  🛡 CodeQL Initialize   ◄── still first, but…    │
│  Install Go 1.25                                   │
│  Initialize go workspace  (with version-scoped    │
│                            replace for v0.6.0-    │
│                            preview to avoid       │
│                            ambiguous imports)     │
│  Vet root / submodules                            │
│  Build all Go modules   ◄── single traced script  │
│    (eng/ci/scripts/codeql-build.sh builds root    │
│     + every submodule + samples in one pass)      │
│  🛡 CodeQL Finalize  →  real DB, real findings   │
│                                                   │
└───────────────────────────────────────────────────┘
```

### DIFF
--- a/eng/ci/templates/jobs/build.yml
+++ b/eng/ci/templates/jobs/build.yml
@@ -5,22 +5,18 @@
 #   1. Installs Go and sets up a workspace so all modules build together.
 #   2. Runs `go vet` for basic static analysis.
 #   3. Builds everything (root + submodules + samples) in one script.
-#   4. Confirms CodeQL actually captured the Go source.
 #
-# Why the order matters:
-#   The 1ES pipeline template automatically inserts a CodeQL Init step
-#   at the very start of the job. CodeQL needs Go to already be installed
-#   and the multi-module workspace to already exist when it starts —
-#   otherwise it initializes with the wrong view of the repo and ends
-#   up capturing zero Go code.
+# About step ordering:
+#   The 1ES pipeline template adds a CodeQL setup step at the start
+#   of the job automatically. We would prefer to install Go and set
+#   up the workspace before CodeQL runs, but the way to do that
+#   (templateContext.preSteps) is not honored by our current 1ES
+#   template version.
 #
-#   To make sure Go and go.work are ready before CodeQL Init runs,
-#   we put those setup steps under `templateContext.preSteps`.
-#   Anything under `preSteps` runs before the CodeQL Init step.
-#
-#   The main build then runs from a single script (codeql-build.sh)
-#   between CodeQL Init and CodeQL Finalize. Doing every module's
-#   build in one place is what lets CodeQL see all of them.
+#   As a workaround, Go setup runs as the first normal steps below.
+#   The agent already has an older Go on PATH, so CodeQL still has
+#   something to watch, and later `go` commands should still be
+#   captured.
 # ============================================================
 
 parameters:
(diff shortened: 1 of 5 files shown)


eng/ci/templates/jobs/build.yml
```diff
@@ -23,52 +39,52 @@ jobs:
...
+    #     still contains files now moved into submodules, and
+    #     `go vet`/`go build` fail with "ambiguous import".
+    #     Note: we pin the specific version rather than "all versions"
+    #     because `.` is already in the workspace `use` list, and Go
+    #     rejects an all-versions replace of a module already in use.
+    #   * Second replace: swaps in a newer genproto version that plays
+    #     nicely with the OpenTelemetry collector deps.
     - script: |
         go work init . ./middleware/otelfunc ./otelcollector ./triggers/blob
+        go work edit -replace github.com/azure/azure-functions-golang-worker@v0.6.0-preview=.
```
ahmedmuhsin: Is there a way we can derive the go worker version instead of hardcoding it so that we don't have to update the CI every time we release a new version?

eng/ci/templates/jobs/unit-tests.yml
```diff
@@ -12,15 +30,28 @@ jobs:
...
 
-    # See eng/ci/templates/jobs/build.yml for rationale.
+    # Same workspace setup as the Build job. See eng/ci/templates/jobs/build.yml
+    # for why this workspace + genproto replace is needed.
     - script: |
         go work init . ./middleware/otelfunc ./otelcollector ./triggers/blob
+        # Force Go to use the local checkout as the root module instead
+        # of downloading an older published copy. See build.yml for the
+        # full explanation.
+        go work edit -replace github.com/azure/azure-functions-golang-worker@v0.6.0-preview=.
```
ahmedmuhsin: Is there a way we can derive the go worker version instead of hardcoding it so that we don't have to update the CI every time we release a new version?

vrdmr (approved):

eng/ci/templates/jobs/build.yml
harshivcodes: addressed your comment - watching the build to make sure it passes!

vrdmr (approved):

ahmedmuhsin (approved):

PR #55, opened 2026-07-28
worker: wait for gRPC READY before opening EventStream

harshivcodes:
**What**
- Add shared `internal/hostclient` connection setup for both the worker and proxy.
- Drive lazy gRPC connections to `connectivity.Ready` before opening `EventStream`.
- Use a 50-second readiness timeout, leaving headroom within the host’s 60-second startup budget for stream setup and error logging.
- Configure faster reconnect backoff:
  - 100ms base delay
  - 1.6 multiplier
  - 20% jitter
  - 500ms maximum delay
  - 1-second minimum connect timeout
- Fail immediately if the connection reaches terminal `Shutdown`.
- Close the connection when readiness or stream creation fails.
- Preserve wrapped errors so callers can inspect failures with `errors.Is`.
- Apply the same behavior to the Flex proxy while retaining the worker- and proxy-specific `StartStream` messages.
- File #56 to track the pre-existing `startSender` queue-draining issue separately.

**Why**
`grpc.NewClient` creates a lazy connection. Opening `EventStream` immediately means a single transient connection failure can terminate the worker or proxy before gRPC reconnects.

Waiting for `READY` allows gRPC to retry within the host startup budget. The tuned backoff avoids long gaps between connection attempts, while the 50-second deadline leaves time for the timeout diagnostic to reach the logs before the host’s startup timeout expires.

**Tests**
- Verify the connection policy and 50-second timeout.
- Cover already-ready and state-transition paths.
- Verify `Connect()` activates the lazy connection.
- Verify timeouts include the last observed state and wrap `context.DeadlineExceeded`.
- Verify terminal `Shutdown` fails immediately.
- Verify failed readiness and stream creation close the connection.
- Verify worker error wrapping and proxy-specific `StartStream` behavior.

### DIFF
--- a/worker/grpc.go
+++ b/worker/grpc.go
@@ -7,24 +7,23 @@ import (
 	"log/slog"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	pb "github.com/azure/azure-functions-golang-worker/worker/proto"
 
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/connectivity"
 	"google.golang.org/grpc/credentials/insecure"
 )
 
-// outboundQueueSize is the buffered size of the channel used by the gRPC
-// sender goroutine. Sized for typical Functions workloads where a single
-// invocation might emit a handful of logs plus the response message; the
-// host drains messages quickly enough that backpressure is rare.
+// outboundQueueSize is the capacity, in messages, of the buffered channel
+// feeding the gRPC sender goroutine. Each slot holds one *pb.StreamingMessage
+// (a response or a log line). Sized generously; the host drains fast enough
+// that backpressure is rare.
 const outboundQueueSize = 256
 
-// streamSender is a goroutine-safe push onto the worker's outbound gRPC
-// queue. Returned by [startSender]; passed wherever a response or log
-// message needs to leave the worker. The signature matches what
-// [log.NewWriter] accepts as its send parameter, so the worker can wire
-// the same function into both the dispatcher and the log writer.
+// streamSender enqueues a message onto the worker's outbound gRPC queue.
+// Safe for concurrent use. Its signature matches what [log.NewWriter] expects.
 type streamSender func(*pb.StreamingMessage) error
 
 func connectToHost(hostAddress string, maxMsgSize int, workerId string) (
@@ -51,12 +50,54 @@ func getBidiStreamClient(address string, maxMsgSize int) (grpc.BidiStreamingClie
 		return nil, err
 	}
 
+	// Drive the connection to READY before opening the stream. Otherwise
+	// EventStream would fast-fail if the host isn't listening yet, which
+	// happens when the host launches the worker as part of its startup.
+	if err := waitForConnReady(conn, connectTimeout); err != nil {
+		_ = conn.Close()
+		return nil, err
+	}
+
 	client := pb.NewFunctionRpcClient(conn)
 	return client.EventStream(context.Background())
 }
 
-// sendStartStreamMessage establishes the worker -> host handshake. After
-// it returns, the host sends WorkerInitRequest.
+// connectTimeout bounds the initial wait for the host's gRPC server. If it
+// isn't listening within this window, fail loudly instead of hanging.
+const connectTimeout = 5 * time.Second
+
+// waitForConnReady blocks until conn reaches connectivity.Ready or the
+// timeout expires. gRPC handles dial retries and backoff internally.
+func waitForConnReady(conn readyConn, timeout time.Duration) error {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	// grpc.NewClient is lazy: the connection stays in IDLE until the first
+	// RPC. Without this call the loop below has nothing to observe and
+	// every startup would hang for the full connectTimeout.
+	conn.Connect()
+	for {
+		state := conn.GetState()
+		if state == connectivity.Ready {
+			return nil
+		}
+		if !conn.WaitForStateChange(ctx, state) {
+			return fmt.Errorf("timed out after %s waiting for gRPC connection to become ready (last state: %s): %w",
+				timeout, state, ctx.Err())
+		}
+	}
+}
+
+// readyConn is the subset of *grpc.ClientConn used by waitForConnReady,
+// carved out so tests can supply a fake.
+type readyConn interface {
+	Connect()
+	GetState() connectivity.State
+	WaitForStateChange(ctx context.Context, sourceState connectivity.State) bool
+}
+
+// sendStartStreamMessage performs the worker -> host handshake, after which
+// the host sends WorkerInitRequest.
 func sendStartStreamMessage(client grpc.BidiStreamingClient[pb.StreamingMessage, pb.StreamingMessage], workerId string) error {
 	startStreamMsg := &pb.StreamingMessage{
 		Content: &pb.StreamingMessage_StartStream{
@@ -68,39 +109,32 @@ func sendStartStreamMessage(client grpc.BidiStreamingClient[pb.StreamingMessage,
 	return client.Send(startStreamMsg)
 }
 
-// startSender owns the gRPC ClientStream's Send call. gRPC's Send is not
-// safe for concurrent use, so we serialize all writes through a single
-// goroutine that drains a buffered channel.
+// startSender serializes writes to the gRPC stream. grpc.ClientStream.Send
+// is not safe for concurrent use, so every message goes through a single
+// goroutine draining a buffered channel.
 //
 // Returns:
-//   - send: a goroutine-safe enqueue function. Pass it to LogWriter and
-//     anywhere a response message needs to be emitted.
-//   - stop: shuts the sender goroutine down. Safe to call any number of
-//     times concurrently; only the first call closes the queue.
-//   - done: closed when the sender goroutine has exited (either because
-//     stop was called or because Send returned an error). Callers may
-//     wait on it to know the gRPC stream is no longer usable.
+//   - send: enqueues a message; safe for concurrent callers.
+//   - stop: shuts the sender down. Idempotent.
+//   - done: closed once the sender goroutine has exited (via stop or a
+//     Send error). Signals that the gRPC stream is no longer usable.
 //
-// When client.Send returns an error the sender logs it through the
-// supplied bootstrap handler (so we don't deadlock by recursing through
-// the LogWriter) and exits. Subsequent enqueues become no-ops returning
-// the captured error.
+// On Send failure the sender logs via errLogger — not the LogWriter, which
+// would recurse back through this same sender — and exits. Later enqueues
+// return io.ErrClosedPipe.
 func startSender(client grpc.BidiStreamingClient[pb.StreamingMessage, pb.StreamingMessage], errLogger *slog.Logger) (send streamSender, stop func(), done <-chan struct{}) {
 	queue := make(chan *pb.StreamingMessage, outboundQueueSize)
 	finished := make(chan struct{})
 
-	// stop is a close-once channel that signals shutdown to all senders.
-	// It allows multiple concurrent readers (sendFn calls) to detect shutdown
-	// without races. The sender goroutine drains the queue after stop is
-	// closed, ensuring no messages are lost.
+	// stopChan is close-once so multiple sendFn callers can detect shutdown
+	// without racing. closed is a fast-path check that avoids a channel op
+	// on the hot enqueue path.
 	stopChan := make(chan struct{})
 	var stopOnce sync.Once
 	var closed atomic.Bool
 
 	go func() {
 		defer close(finished)
-		// Sender goroutine: dequeue and forward messages until stopChan closes
-		// or a gRPC send error occurs.
 		for {
 			select {
 			case <-stopChan:
@@ -140,7 +174,6 @@ func startSender(client grpc.BidiStreamingClient[pb.StreamingMessage, pb.Streami
 		}
 		select {
 		case <-stopChan:
-			// Shutdown has started; return immediately without queuing.
 			return io.ErrClosedPipe
 		case queue <- m:
 			return nil
--- a/worker/grpc.go
+++ b/worker/grpc.go
@@ -81,6 +81,14 @@ func waitForConnReady(conn readyConn, timeout time.Duration) error {
 		if state == connectivity.Ready {
 			return nil
 		}
+		// Shutdown is terminal — the channel can never become Ready again,
+		// so waiting for the full timeout would only add startup latency
+		// and hide the real failure. Idle/Connecting/TransientFailure are
+		// all recoverable; gRPC drives its own retries and backoff between
+		// them, so we just keep observing.
+		if state == connectivity.Shutdown {
+			return fmt.Errorf("gRPC connection is in terminal state %s and cannot become ready", state)
+		}
 		if !conn.WaitForStateChange(ctx, state) {
 			return fmt.Errorf("timed out after %s waiting for gRPC connection to become ready (last state: %s): %w",
 				timeout, state, ctx.Err())
--- a/worker/grpc.go
+++ b/worker/grpc.go
@@ -7,13 +7,11 @@ import (
 	"log/slog"
 	"sync"
 	"sync/atomic"
-	"time"
 
+	"github.com/azure/azure-functions-golang-worker/internal/hostclient"
 	pb "github.com/azure/azure-functions-golang-worker/worker/proto"
 
 	"google.golang.org/grpc"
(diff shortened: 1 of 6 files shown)


worker/grpc.go
```diff
@@ -51,12 +50,54 @@ func getBidiStreamClient(address string, maxMsgSize int) (grpc.BidiStreamingClie
...
+
+	// grpc.NewClient is lazy: the connection stays in IDLE until the first
+	// RPC. Without this call the loop below has nothing to observe and
+	// every startup would hang for the full connectTimeout.
+	conn.Connect()
+	for {
+		state := conn.GetState()
+		if state == connectivity.Ready {
+			return nil
+		}
```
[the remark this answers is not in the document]
gavin-aguiar: Do we also need to handle these states:

``` go
	// Idle indicates the ClientConn is idle.
	Idle [State](https://pkg.go.dev/google.golang.org/grpc/connectivity#State) = [iota](https://pkg.go.dev/builtin#iota)
	// Connecting indicates the ClientConn is connecting.
	Connecting
	// Ready indicates the ClientConn is ready for work.
	Ready
	// TransientFailure indicates the ClientConn has seen a failure but expects to recover.
	TransientFailure
	// Shutdown indicates the ClientConn has started shutting down.
	Shutdown
```
Ref: https://pkg.go.dev/google.golang.org/grpc/connectivity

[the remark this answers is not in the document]
harshivcodes: the conn.Connect() nudges the connection from Idle state - so that's already handled. Connecting, TransientFailure are transitional states that will reach a terminal state - the comment about unhandled Shutdown is valid. adding coverage for that!

worker/grpc_test.go
```diff
@@ -97,3 +99,113 @@ func TestStartSender_SendErrorClosesDoneAndRejectsFutureSends(t *testing.T) {
 		t.Fatalf("send after sender failure: got %v want %v", err, io.ErrClosedPipe)
 	}
 }
+
+// fakeReadyConn drives waitForConnReady in tests without needing a real gRPC
+// client. States are consumed in order: each GetState() pop reflects the next
+// scripted state, and WaitForStateChange blocks (respecting ctx) until either
+// another state is pushed or the deadline expires. This models gRPC's own
+// state-notification semantics closely enough for the retry-timing tests.
```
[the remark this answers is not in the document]
harshivcodes: !

worker/grpc_test.go
```diff
@@ -97,3 +99,113 @@ func TestStartSender_SendErrorClosesDoneAndRejectsFutureSends(t *testing.T) {
...
+
+	if err == nil {
+		t.Fatal("expected timeout error, got nil")
+	}
+	if !strings.Contains(err.Error(), "timed out") {
+		t.Fatalf("error should mention timeout, got: %v", err)
+	}
+	if !errors.Is(err, context.DeadlineExceeded) {
+		t.Fatalf("error should wrap context.DeadlineExceeded, got: %v", err)
+	}
```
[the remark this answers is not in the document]
harshivcodes: fixed the comments

worker/grpc.go
```diff
@@ -51,12 +50,62 @@ func getBidiStreamClient(address string, maxMsgSize int) (grpc.BidiStreamingClie
 		return nil, err
 	}
 
+	// Drive the connection to READY before opening the stream. Otherwise
+	// EventStream would fast-fail if the host isn't listening yet, which
+	// happens when the host launches the worker as part of its startup.
```
ahmedmuhsin: I went looking for this race in the host and couldn't find it. `RpcInitializationService.StartAsync` awaits `InitializeRpcServerAsync` before `InitializeChannelsAsync`, and `AspNetCoreGrpcServer.StartAsync` awaits `_grpcHost.StartAsync`, so Kestrel is bound before any worker process gets launched. The proxy path is the same, `startLocalServer` runs before we exec the child.

The fast fail is still worth fixing, I just think the trigger is a transient connect failure rather than the host not listening.

What did you see that led to this? If there's a failed run or a repro, let's link it. Otherwise let's reword the comment to say we don't want a single failed connect to take the process down.

worker/grpc.go
```diff
@@ -51,12 +50,62 @@ func getBidiStreamClient(address string, maxMsgSize int) (grpc.BidiStreamingClie
...
+	if err := waitForConnReady(conn, connectTimeout); err != nil {
```
ahmedmuhsin: The 5s budget only buys us about 2.6s of real coverage. gRPC's default reconnect backoff is BaseDelay 1s with Multiplier 1.6, so connect attempts land at roughly t=0, 1.0s, and 2.6s. The next attempt is at 5.2s, which is after we've already given up. Nothing gets retried between 2.6s and the deadline, so the back half of the window is dead time.

Stub server that starts listening at 3.5s, 6 trials each.

| backoff | connected |
| --- | --- |
| default | 2 of 6, the other 4 hit the timeout and would have exited the worker |
| tuned below | 6 of 6, each within ~350ms of the server coming up |

Let's add connect params to `opts` above.

```go
grpc.WithConnectParams(grpc.ConnectParams{
	Backoff:           backoff.Config{BaseDelay: 100 * time.Millisecond, Multiplier: 1.6, Jitter: 0.2, MaxDelay: 500 * time.Millisecond},
	MinConnectTimeout: time.Second,
})
```

worker/grpc.go
```diff
@@ -51,12 +50,62 @@ func getBidiStreamClient(address string, maxMsgSize int) (grpc.BidiStreamingClie
...
+
 	client := pb.NewFunctionRpcClient(conn)
 	return client.EventStream(context.Background())
 }
 
-// sendStartStreamMessage establishes the worker -> host handshake. After
-// it returns, the host sends WorkerInitRequest.
+// connectTimeout bounds the initial wait for the host's gRPC server. If it
+// isn't listening within this window, fail loudly instead of hanging.
+const connectTimeout = 5 * time.Second
```
ahmedmuhsin: 5s is a lot tighter than what the platform gives us. `ProcessStartupTimeout` is 60s and covers launch through StartStream, `ProcessRestartInterval` is 10s so a false timeout costs another 10s of cold start, and the Flex native worker.config.json sets `initializationTimeout` to 2 minutes.

Python being at 5s isn't a strong argument since it has the same backoff gap we do.

Can we raise this closer to the host budget, or read it from an env var so we can tune it without a release?

worker/grpc.go
```diff
@@ -51,12 +50,62 @@ func getBidiStreamClient(address string, maxMsgSize int) (grpc.BidiStreamingClie
...
+		state := conn.GetState()
+		if state == connectivity.Ready {
+			return nil
+		}
+		// Shutdown is terminal — the channel can never become Ready again,
+		// so waiting for the full timeout would only add startup latency
+		// and hide the real failure. Idle/Connecting/TransientFailure are
+		// all recoverable; gRPC drives its own retries and backoff between
+		// them, so we just keep observing.
+		if state == connectivity.Shutdown {
```
ahmedmuhsin: Small one. This can't trigger on the only call path, `conn` is created a few lines up and nothing closes it before the wait. Fine to keep as a guard, let's just note it's for future callers so nobody goes hunting for the case that hits it.

worker/grpc.go
```diff
@@ -51,12 +50,62 @@ func getBidiStreamClient(address string, maxMsgSize int) (grpc.BidiStreamingClie
...
+			return fmt.Errorf("gRPC connection is in terminal state %s and cannot become ready", state)
+		}
+		if !conn.WaitForStateChange(ctx, state) {
+			return fmt.Errorf("timed out after %s waiting for gRPC connection to become ready (last state: %s): %w",
```
ahmedmuhsin: We wrap with `%w` here and the new test asserts `errors.Is(err, context.DeadlineExceeded)`, but `connectToHost` above re-wraps with `%v` and drops the chain, so nothing past that function can match on it. Let's switch those two wraps to `%w` as well.

worker/grpc.go
```diff
@@ -68,39 +117,32 @@ func sendStartStreamMessage(client grpc.BidiStreamingClient[pb.StreamingMessage,
...
 	var stopOnce sync.Once
 	var closed atomic.Bool
 
 	go func() {
 		defer close(finished)
-		// Sender goroutine: dequeue and forward messages until stopChan closes
-		// or a gRPC send error occurs.
 		for {
 			select {
 			case <-stopChan:
```
ahmedmuhsin, line 89: The comment removed in this hunk said the goroutine drains the queue after stop so no messages are lost. That was wrong, and it was also the only thing pointing at a real bug.

When `stop()` runs and the queue isn't empty, both cases in this select are ready and Go picks one at random, so the goroutine usually returns and drops what's left. I probed it on this branch with 50 queued messages and a `Send` that takes 2ms. 50 accepted, 2 to 4 actually sent, the rest dropped, same across 5 runs.

This predates your change so I don't want to block on it. Can we file a follow up so it doesn't get lost now that the comment is gone? It shows up as missing logs on shutdown.

worker/grpc_test.go
```diff
@@ -97,3 +99,158 @@ func TestStartSender_SendErrorClosesDoneAndRejectsFutureSends(t *testing.T) {
...
+// client. GetState() returns the most recently pushed state (matching gRPC's
+// "current state" semantics — nothing is consumed). WaitForStateChange blocks,
+// respecting ctx, until a state different from sourceState has been pushed or
+// the deadline expires. This models gRPC's own state-notification semantics
+// closely enough for the retry-timing tests.
+type fakeReadyConn struct {
+	mu             sync.Mutex
+	states         []connectivity.State
+	changeCh       chan struct{}
+	connectCalled  bool
```
ahmedmuhsin: gofmt wants one space before `bool` here and the other three fields shifted left by one.

worker/grpc_test.go
```diff
@@ -97,3 +99,158 @@ func TestStartSender_SendErrorClosesDoneAndRejectsFutureSends(t *testing.T) {
...
+	}()
+
+	start := time.Now()
+	if err := waitForConnReady(conn, 2*time.Second); err != nil {
+		t.Fatalf("expected nil, got %v", err)
+	}
+	if elapsed := time.Since(start); elapsed > time.Second {
+		t.Fatalf("waited too long for ready state: %s", elapsed)
+	}
+	if !conn.connectCalled {
```
ahmedmuhsin: `connectCalled` is written under `f.mu` but read without it here and in the other two tests. It's only safe because `Connect` runs on the test goroutine. Let's make it an `atomic.Bool`.

worker/grpc_test.go
```diff
@@ -97,3 +99,158 @@ func TestStartSender_SendErrorClosesDoneAndRejectsFutureSends(t *testing.T) {
...
+func TestWaitForConnReady_AlreadyReady(t *testing.T) {
+	// If the channel is already Ready when we check, waitForConnReady
+	// should return immediately without waiting for any state change.
+	conn := newFakeReadyConn(connectivity.Ready)
+
+	start := time.Now()
+	if err := waitForConnReady(conn, 2*time.Second); err != nil {
+		t.Fatalf("expected nil, got %v", err)
+	}
+	if elapsed := time.Since(start); elapsed > 50*time.Millisecond {
```
ahmedmuhsin: 50ms is tight for a wall clock assert. This is the one I'd expect to flake first on a busy agent, can we loosen it?

ahmedmuhsin (commented):
Ran this against a stub host. The fast fail is there, a dial against a dead port comes back `Unavailable` in about 2ms, so today one failed connect takes the worker down.

Two things I'd like to sort out before this goes in. The 5s budget only gives us about 2.6s of real coverage, and I couldn't reproduce the startup race the PR describes. Both inline.

proxy/main.go has the same lazy dial into `EventStream` with no readiness wait, and on Flex the proxy is the process the host launches. Can we fix both here or file a follow up for the proxy?

worker/grpc.go
harshivcodes: Good catch, fixed. Both wraps in `connectToHost` now use `%w`:

```go
return nil, fmt.Errorf("failed to connect to gRPC stream: %w", err)
...
return nil, fmt.Errorf("failed to send start stream message: %w", err)
```

So the chain from `waitForConnReady` -> `connectToHost` -> caller is preserved end-to-end, and `errors.Is(err, context.DeadlineExceeded)` matches at any layer, not just inside `waitForConnReady`.

worker/grpc.go
harshivcodes: Fair point, you're right that Kestrel is bound before the child process launches so it's not really a race with the listener. Reworded the comment on `waitForConnReady` (and the PR description) to frame it as "don't let a single transient connect failure take the worker down" instead of a host-listener race. No repro run to link — this came out of reading the Python worker's 5s timeout and wanting parity, not a specific failed job.

worker/grpc.go
harshivcodes: Nice analysis, that backoff-gap math is exactly the failure mode I didn't want to leave on the table. Added the `WithConnectParams` block to `opts` in `getBidiStreamClient` with the values you suggested (BaseDelay 100ms, Multiplier 1.6, Jitter 0.2, MaxDelay 500ms, MinConnectTimeout 1s). With that, retries land at ~0, 100ms, 260ms, 516ms, then capped at 500ms, so the 5s window gets ~9 attempts instead of 3 and the back half of the window is actually useful.

worker/grpc.go
harshivcodes: Fair — bumped the default to 60s to match `ProcessStartupTimeout`, and made it overridable via `FUNCTIONS_WORKER_GRPC_CONNECT_TIMEOUT` (duration string) so we can tune without a release.

worker/grpc.go
harshivcodes: Yeah, on the only current call path `conn` was just created and nothing closes it before the wait, so this branch is dead today. Kept it as a guard for future callers and added a comment on the `if state == connectivity.Shutdown` block saying so, so nobody goes hunting for the reproducer.

worker/grpc.go
harshivcodes: Good find, and thanks for the probe numbers — that's a real bug, not just a stale comment. Tracked in #56 (drain the queue on stop, bounded by a small deadline so shutdown still terminates) so it doesn't get lost. Agreed on not blocking this PR on it since it predates the change.

worker/grpc_test.go
harshivcodes: Done.

worker/grpc_test.go
harshivcodes: Right, it's only accidentally safe because `Connect` runs on the test goroutine before the reader looks. Switched `connectCalled` to `atomic.Bool` — `Connect()` does a `Store(true)` (no more `f.mu` for that field) and the test assertions use `Load()`. `-race` is clean.

worker/grpc_test.go
harshivcodes: bumped it to 500ms!

worker/grpc.go
ahmedmuhsin: Checked for this one and I don't think it made it in. `connectionTimeout` in hostclient.go is still a plain const and there's no `os.Getenv` for it anywhere on the branch. I also grepped the host for `GRPC_CONNECT_TIMEOUT` in case it was a host side setting being passed down and there's nothing there either, and Python's `grpc_connect_timeout` is a constructor arg rather than an env var. Did that change get dropped before the push?

Separate thing on the 60s. `ProcessStartupTimeout` covers launch all the way through StartStream, so if we spend the full 60s waiting for READY the host kills us right as we give up and our timeout error never makes it to the log. Can we set it a little under, say 50s, so the "timed out after 50s waiting for gRPC connection to become ready" message actually lands?

harshivcodes: The env-var change was not dropped; we decided to keep this as a fixed timeout for now and add configurability only if telemetry or reported failures show it is needed.

The 60s diagnostic race is valid. I have changed the timeout to 50s locally, leaving roughly 10s of the host's `ProcessStartupTimeout` budget for `EventStream`/`StartStream` setup and for the timeout error to reach the logs. That update has not been pushed yet.

ahmedmuhsin (approved):
The PR description is still the original text and out of date on most points.

It says 5s matching Python, it's 50s. It puts `waitForConnReady` and `readyConn` in worker/grpc.go, both moved to internal/hostclient as `waitForReady` and `readyConnection`. The tests section points at worker/grpc_test.go and `fakeReadyConn`, now internal/hostclient/hostclient_test.go and `fakeConnection`. Nothing about the shared package, the proxy moving onto it, or the backoff change.

The Why section still says opening `EventStream` would race the host's listener, which we agreed to drop. On a squash merge that becomes the commit message.

Can we refresh it before merging?

gavin-aguiar (approved):

PR #59, opened 2026-08-04
test: add reliable local integration test harness and runner

harshivcodes:
Integration tests were previously runnable only through manual local setup. This PR is the first step toward making them a required Azure DevOps check for every PR.

It adds a reliable local test runner, cross-platform Functions host lifecycle management, emulator readiness checks, automatic diagnostics, and cleanup. The HTTP integration test is the first scenario migrated to the new harness.

Follow up PRs will add additional scenarios.

**Run tests locally**
* From the repository root:
```powershell
Set-Location tests\integration
go run ./cmd/integrationtest
```

| File | Purpose |
| --- | --- |
| `tests/integration/internal/testhost/host.go` | Functions host lifecycle, readiness, logging, and cleanup |
| `tests/integration/internal/testhost/host_test.go` | Host lifecycle regression tests |
| `tests/integration/internal/testrunner/runner.go` | Test orchestration, Azurite readiness, artifacts, and cleanup |
| `tests/integration/internal/testrunner/runner_test.go` | Runner orchestration unit tests |
| `tests/integration/cmd/integrationtest/main.go` | Shared integration-test command |
| `tests/integration/http_test.go` | First test migrated to the new harness |
| `tests/integration/README.md` | Target workflow and Azure DevOps integration contract |


tests/integration/cmd/integrationtest/main_test.go
```diff
@@ -0,0 +1,27 @@
...
+	if runner.ArtifactDir != "artifacts" {
+		t.Fatalf("ArtifactDir = %q", runner.ArtifactDir)
+	}
+	if runner.TestPattern != "^TestHttpTriggerGet$" {
+		t.Fatalf("TestPattern = %q", runner.TestPattern)
+	}
+	if runner.FuncExe != "func" {
+		t.Fatalf("FuncExe = %q", runner.FuncExe)
+	}
+	if runner.CoreToolsVersion != "4.12.0" {
```
gavin-aguiar: Why do we need a version check? Can we always just use the latest?

harshivcodes: we can - go support is technically only available in 4.12+ as long as we configure the pipeline with the correct version, we should be good to go.

tests/integration/cmd/integrationtest/main.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/azure/azure-functions-golang-worker/tests/integration/internal/testrunner"
+)
+
+func main() {
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
```
harshivcodes: agent: use a variable for this; also add a brief comment that this duration is used for the whole test suite

tests/integration/cmd/integrationtest/main_test.go
harshivcodes: agent: update the check to verify core tools version >= 4.12

tests/integration/README.md
```diff
@@ -1,116 +1,241 @@
...
+Diagnostics from the runner, emulator, host, and worker
+are written to the artifacts directory.
+```
 
 ## Prerequisites
 
 1. **Go 1.25+**
-2. **Azure Functions Core Tools** (`func.exe`) — built or installed
-3. **Docker** — for running emulators
+2. **Azure Functions Core Tools 4.12.0**
```
harshivcodes: agent: should be > 4.12

tests/integration/blob_test.go
```diff
@@ -19,7 +19,6 @@ const azuriteConnStr = "DefaultEndpointsProtocol=http;" +
 
 var blobEnv = map[string]string{
 	"AzureWebJobsStorage":      azuriteConnStr,
```
ahmedmuhsin: gofmt wants the padding dropped now that `FUNCTIONS_WORKER_RUNTIME` is gone, the value column was sized to that longer key.

```go
"AzureWebJobsStorage": azuriteConnStr,
```

Same in cosmosdb_test.go, eventgrid_test.go, and eventhub_test.go. queue_storage_test.go and timer_test.go already got it right, so `gofmt -w` on the four that were missed covers it.

tests/integration/internal/testrunner/runner.go
```diff
@@ -0,0 +1,270 @@
...
+	// Do not start Core Tools until the emulator responds at the service layer.
+	// A running container or open TCP port alone does not prove readiness.
+	if err := waitReady(ctx); err != nil {
+		return fmt.Errorf("wait for Azurite readiness: %w", err)
+	}
+
+	// Keep Go test execution as a child process so local and CI runs use the same
+	// test selection, timeout, output, and exit status.
+	testArgs := []string{"test", "-v", "-count=1", "-timeout", "120s", "-run", r.TestPattern, "."}
+	testOutput, err := command(ctx, "go", testArgs...)
```
ahmedmuhsin: `CombinedOutput` buffers the whole run, so `go test -v` prints nothing until it finishes, and nothing at all if the suite timeout fires first. That's livable locally but once this is a required check the ADO log stays empty for the entire run and a timeout gives us no output to debug from.

Can we stream and keep the artifact, something like an `io.MultiWriter` over `os.Stdout` and a buffer for the test command?

tests/integration/cmd/integrationtest/main_test.go
harshivcodes: Done

tests/integration/blob_test.go
harshivcodes: Fixed formatting issues.

tests/integration/internal/testrunner/runner.go
harshivcodes: Done — test output now streams live to the local console or Azure DevOps log and directly to `go-test.log`, without buffering the full run in memory.

ahmedmuhsin (approved):

gavin-aguiar (approved):

PR #68, opened 2026-08-12
Add strongly typed Event Hub and Service Bus batch triggers

### LINKED ISSUE #46: [Go]: Event Hub and Service Bus triggers can't receive batches (cardinality "many")
Right now the Event Hub and Service Bus triggers only hand the function one message per invocation. The handler aliases are single message.

```go
type EventHubHandler   = func(context.Context, bindings.EventHubMessage) error
type ServiceBusHandler = func(context.Context, bindings.ServiceBusMessage) error
```

On the surface it looks like batch is already supported. Both bindings carry a `Cardinality` field and there's a `WithCardinality` option, so you'd expect `WithCardinality("many")` to give you a batch. It doesn't. `app.EventHub`, `app.ServiceBusQueue`, and `app.ServiceBusTopic` all hardcode `Cardinality: "one"`, and even when you flip it to `"many"` the handler still only accepts a single message struct, so a delivered batch has nowhere to land. The option changes the binding JSON and nothing else.

The host side is already in good shape. The worker advertises `TypedDataCollection` in `handleWorkerInitRequest`, so the host is willing to send a collection when cardinality is many. The missing piece is entirely worker side, a slice typed handler plus the converter path that fills it.

We already have the shape for this in the same codebase. Cosmos DB and SQL both deliver batches today and decode into a slice.

```go
type CosmosDBHandler  = func(context.Context, []bindings.CosmosDocument) error
type SQLChangeHandler = func(context.Context, []bindings.SQLChange) error
```

So the fix should follow how those two already work.

What I think we need

- Accept `[]bindings.EventHubMessage` and `[]bindings.ServiceBusMessage` handler signatures
- Teach the converter to populate the slice when the host delivers a collection
- Make `WithCardinality("many")` actually work with the slice handler. Until it does, either guard it or document that single message is the only mode that works, since today it silently does nothing useful

For reference, Storage Queue (the new `queueTrigger` in #45) is out of scope. Storage queues don't batch by design, each message is its own invocation, so a single message handler is correct there. This is specifically about Event Hub and Service Bus, where batch is a real feature the other language workers expose and we don't.

harshivcodes:
## Summary

- Add `EventHubBatch`, `ServiceBusQueueBatch`, and `ServiceBusTopicBatch` APIs with strongly typed handlers.
- Decode typed-data collections while preserving empty entries and per-message metadata alignment.
- Configure Event Hubs and Service Bus trigger payloads as binary for both single-message and batch handlers, preventing non-UTF-8 payload corruption.
- Update samples, emulator configuration, documentation, and integration coverage.

## Breaking change and release notes

- **Cosmos DB trigger fix:** `CosmosDocument.Timestamp` changes from `string` to `int64`. Cosmos DB emits `_ts` as a JSON number, so the previous public type caused real change-feed documents to fail decoding on both JSON and string input paths. Consumers assigning or comparing `Timestamp` as a string must update their code.
- **Event Hubs and Service Bus payload fix:** trigger bindings now request binary data. Existing single-message handlers, as well as new batch handlers, receive the original bytes instead of UTF-8 replacement bytes for protobuf, Avro, compressed, or other non-text payloads.
- Queue, Service Bus, and Event Hub message bodies consistently marshal under the public JSON property `body`; the worker-specific raw-data marker is carried by `azfunc:"data"`.

## Testing

- Added unit tests for batch conversion, empty-entry alignment, metadata precedence, malformed extension metadata, bindings, and handler invocation.
- Expanded Event Hub and Service Bus integration tests for single and batch delivery.
- The checked-in integration runner remains scoped to the HTTP/Azurite milestone; Event Hubs, Service Bus, and Cosmos DB integration tests run when their emulator stack is started explicitly as documented in `tests/integration/README.md`.

Fixes #46

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,17 @@ app.SQL("productsChanged", productsChanged,
     sdk.WithTable("dbo.Products"),
     sdk.WithConnection("AzureWebJobsSqlConnectionString"),
 )
+
+// Event Hub and Service Bus expose separate, strongly typed batch methods.
+app.EventHubBatch("processEvents", eventBatchHandler,
+    sdk.WithEventHubName("events"),
+    sdk.WithConnection("EventHubConnection"),
+)
+
+app.ServiceBusQueueBatch("processMessages", messageBatchHandler,
+    sdk.WithQueueName("messages"),
+    sdk.WithConnection("ServiceBusConnection"),
+)
 ```
 
 ### Extension Triggers (`triggers/`)
--- a/README.md
+++ b/README.md
@@ -169,7 +169,7 @@ For work that matters (processing events, calling APIs), use [`sdk.RecoverTo`](s
 ```go
 import "golang.org/x/sync/errgroup"
 
-func EventHubHandler(ctx context.Context, events []bindings.EventHubEvent) error {
+func EventHubHandler(ctx context.Context, events []bindings.EventHubMessage) error {
     g, ctx := errgroup.WithContext(ctx)
     for _, e := range events {
         g.Go(func() (err error) {
--- a/sdk/bindings/cosmos.go
+++ b/sdk/bindings/cosmos.go
@@ -36,7 +36,7 @@ type CosmosDocument struct {
 	Self        string          `json:"_self"`
 	Etag        string          `json:"_etag"`
 	Attachments string          `json:"_attachments"`
-	Timestamp   string          `json:"_ts"`
+	Timestamp   int64           `json:"_ts"`
 	Lsn         int             `json:"_lsn"`
 }
 
--- a/sdk/bindings/eventhub.go
+++ b/sdk/bindings/eventhub.go
@@ -11,6 +11,7 @@ type EventHubBinding struct {
 	Connection    string `json:"connection"`
 	ConsumerGroup string `json:"consumerGroup,omitempty"`
 	Cardinality   string `json:"cardinality,omitempty"`
+	DataType      string `json:"dataType"`
 }
 
 // EventHubMessage represents a message received from an Azure Event Hub.
@@ -18,7 +19,7 @@ type EventHubBinding struct {
 // Other fields (SequenceNumber, Offset, etc.) are also populated from
 // trigger metadata using case-insensitive matching on their json tags.
 type EventHubMessage struct {
-	Body             json.RawMessage `json:"body"`
+	Body             json.RawMessage `json:"body" azfunc:"data"`
 	EnqueuedTimeUtc  string          `json:"enqueuedTimeUtc"`
 	SequenceNumber   int64           `json:"sequenceNumber"`
 	Offset           string          `json:"offset"`
@@ -50,6 +51,7 @@ func (e *EventHubTrigger) ToBinding() Binding {
 			Connection:    e.Connection,
 			ConsumerGroup: e.ConsumerGroup,
 			Cardinality:   e.Cardinality,
+			DataType:      "binary", // Preserve event bodies as bytes for both single events and batches.
 		},
 	}
 }
--- a/sdk/bindings/eventhub.go
+++ b/sdk/bindings/eventhub.go
@@ -15,7 +15,7 @@ type EventHubBinding struct {
 }
 
 // EventHubMessage represents a message received from an Azure Event Hub.
-// The Body field is populated from the "body" key in trigger metadata.
+// The Body field is populated from the raw trigger input.
 // Other fields (SequenceNumber, Offset, etc.) are also populated from
 // trigger metadata using case-insensitive matching on their json tags.
 type EventHubMessage struct {
--- a/sdk/bindings/servicebus.go
+++ b/sdk/bindings/servicebus.go
@@ -17,12 +17,11 @@ type ServiceBusBinding struct {
 }
 
 // ServiceBusMessage represents a message received from Azure Service Bus.
-// The Body field uses the special json tag "azfuncdata" which instructs the
-// worker's converter to populate this field from the raw trigger input data
-// rather than from trigger metadata. Other fields are populated from trigger
-// metadata using case-insensitive matching on their json tags.
+// The Body field is populated from the raw trigger input.
+// Other fields are populated from trigger metadata using case-insensitive
+// matching on their json tags.
 type ServiceBusMessage struct {
-	Body                    json.RawMessage        `json:"azfuncdata"`
+	Body                    json.RawMessage        `json:"body" azfunc:"data"`
 	ContentType             string                 `json:"contentType"`
 	CorrelationId           string                 `json:"correlationId"`
 	DeliveryCount           int                    `json:"deliveryCount"`
--- a/sdk/bindings/servicebus.go
+++ b/sdk/bindings/servicebus.go
@@ -13,6 +13,7 @@ type ServiceBusBinding struct {
 	Connection        string `json:"connection"`
 	IsSessionsEnabled bool   `json:"isSessionsEnabled,omitempty"`
 	Cardinality       string `json:"cardinality,omitempty"`
+	DataType          string `json:"dataType"`
 }
 
 // ServiceBusMessage represents a message received from Azure Service Bus.
@@ -67,6 +68,7 @@ func (s *ServiceBusQueueTrigger) ToBinding() Binding {
 			Connection:        s.Connection,
 			IsSessionsEnabled: s.IsSessionsEnabled,
 			Cardinality:       s.Cardinality,
+			DataType:          "binary", // Preserve message bodies as bytes for both single messages and batches.
 		},
 	}
 }
@@ -96,6 +98,7 @@ func (s *ServiceBusTopicTrigger) ToBinding() Binding {
 			Connection:        s.Connection,
 			IsSessionsEnabled: s.IsSessionsEnabled,
 			Cardinality:       s.Cardinality,
+			DataType:          "binary", // Preserve message bodies as bytes for both single messages and batches.
 		},
 	}
 }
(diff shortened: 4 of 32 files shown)


sdk/bindings/cosmos.go
```diff
@@ -36,7 +36,7 @@ type CosmosDocument struct {
 	Self        string          `json:"_self"`
 	Etag        string          `json:"_etag"`
 	Attachments string          `json:"_attachments"`
-	Timestamp   string          `json:"_ts"`
+	Timestamp   int64           `json:"_ts"`
```
harshivcodes, line 39: note, this is a bug fix - Cosmos contract is currently broken

worker/converter.go
```diff
@@ -83,6 +84,24 @@ func convertToTypeValue(pt reflect.Type, data *pb.TypedData, tm map[string]*pb.T
 	pv := reflect.New(t)
 	v := pv.Elem()
 
+	if t.Kind() == reflect.Slice && t.Elem().Kind() == reflect.Struct {
```
harshivcodes, line 91: (agent) - add a comment here explaining the code change.

README.md
```diff
@@ -118,6 +118,17 @@ app.SQL("productsChanged", productsChanged,
...
     sdk.WithConnection("AzureWebJobsSqlConnectionString"),
 )
+
+// Event Hub and Service Bus expose separate, strongly typed batch methods.
+app.EventHubBatch("processEvents", eventBatchHandler,
+    sdk.WithEventHubName("events"),
+    sdk.WithConnection("EventHubConnection"),
+)
+
+app.ServiceBusQueueBatch("processMessages", messageBatchHandler,
```
harshivcodes, line 128: exposing separate ServiceBusQueue /  ServiceBusQueueBatch to preserve strong typing for handlers, going this route for better developer experience

gavin-aguiar (approved):

worker/converter.go
```diff
@@ -180,6 +203,81 @@ func convertToTypeValue(pt reflect.Type, data *pb.TypedData, tm map[string]*pb.T
...
+		values := make([]*pb.TypedData, len(collection.GetSint64()))
+		for i, value := range collection.GetSint64() {
+			values[i] = &pb.TypedData{Data: &pb.TypedData_Int{Int: value}}
+		}
+		return values, true
+	}
+	return nil, false
+}
+
+func collectionMetadataAt(metadata map[string]*pb.TypedData, index int) (map[string]*pb.TypedData, error) {
```
ahmedmuhsin, line 241: Two issues here. Neither is reachable with today's Event Hubs and Service Bus extensions (I dumped the real key list, there's no clash), but both are cheap to close.

First, if any key ending in "Array" holds JSON that isn't an array, this returns an error and the entire invocation fails:

```
failed to decode trigger metadata "SomeWeirdArray":
  json: cannot unmarshal object into Go value of type []json.RawMessage
```

We don't own these keys, extensions add new ones over time, and for Service Bus a failing invocation means retry until dead letter. Skipping the key is a much smaller blast radius than failing the batch.

Second, if metadata ever contains both `X` and `XArray`, which one wins comes down to Go map iteration order. I looped a case with both and it flipped on iteration 6 of 200.

Two passes closes both and lets the error return go away:

```go
func collectionMetadataAt(metadata map[string]*pb.TypedData, index int) map[string]*pb.TypedData {
	result := make(map[string]*pb.TypedData, len(metadata))
	for name, value := range metadata {
		if !strings.HasSuffix(strings.ToLower(name), "array") {
			result[name] = value
		}
	}
	// Per-message entries always win over a single-valued twin.
	for name, value := range metadata {
		if !strings.HasSuffix(strings.ToLower(name), "array") {
			continue
		}
		element, ok, err := typedDataElementAt(value, index)
		if err != nil || !ok {
			continue // not a per-message array, skip rather than fail the batch
		}
		result[name[:len(name)-len("Array")]] = element
	}
	return result
}
```

For what it's worth, I checked the rest of this path against a real host and it holds up. Ran the branch against Core Tools 4.12.0 with the Service Bus, Event Hubs and Azurite emulators, all six integration tests pass, and the decoded structs line up per index for Properties, SystemProperties, ApplicationProperties, SequenceNumber, Offset and EnqueuedTimeUtc. The metadata keys the host actually sends match the `*Array` convention this assumes.

sdk/bindings/cosmos.go
ahmedmuhsin, line 39: This deserves more than a bullet at the bottom of the description. On main the Cosmos trigger can't decode a real document at all, because `_ts` is a JSON number and the field was a string:

```
json: cannot unmarshal number into Go struct field CosmosDocument._ts of type string
```

Same on both the `Json` and `String_` paths, so the whole `[]CosmosDocument` decode errors and the invocation fails. This fixes a trigger that was broken end to end, which is a bigger deal than "correct the timestamp type".

It's also a breaking change to a public field, so it should show up in the description and the release notes.

sdk/bindings/servicebus.go
```diff
@@ -67,6 +68,7 @@ func (s *ServiceBusQueueTrigger) ToBinding() Binding {
 			Connection:        s.Connection,
 			IsSessionsEnabled: s.IsSessionsEnabled,
 			Cardinality:       s.Cardinality,
+			DataType:          "binary", // Preserve message bodies as bytes for both single messages and batches.
```
ahmedmuhsin, line 71: Worth spelling out that this changes behaviour for existing single message apps too, not just batches. I sent a non UTF-8 body on both branches:

```
sent      [255 254 0 1 128 65 66]
main      [239 191 189 239 191 189 0 1 239 191 189 65 66]
this PR   [255 254 0 1 128 65 66]
```

Without `dataType` the host hands the body over as a string and every non UTF-8 byte becomes U+FFFD. Anyone pushing protobuf, avro or compressed payloads through these triggers is silently getting corrupted bytes today. That's a real fix, it just needs calling out so people know their bodies change shape.

Related, I confirmed `IncludeEmptyEntriesInMessagePayload` is doing real work rather than being belt and braces. Two message batch where only the second has a ContentType:

```
with the capability:     idx=0 ""                  idx=1 "application/json"
without the capability:  idx=0 "application/json"  idx=1 ""
```

The host drops empty entries from `string[]` metadata arrays, so without it every message after a gap picks up the wrong value. Good catch.

sdk/bindings/eventhub.go
```diff
@@ -11,14 +11,15 @@ type EventHubBinding struct {
...
 }
 
 // EventHubMessage represents a message received from an Azure Event Hub.
-// The Body field is populated from the "body" key in trigger metadata.
+// The Body field is populated from the raw trigger input.
 // Other fields (SequenceNumber, Offset, etc.) are also populated from
 // trigger metadata using case-insensitive matching on their json tags.
 type EventHubMessage struct {
-	Body             json.RawMessage `json:"body"`
+	Body             json.RawMessage `json:"body" azfunc:"data"`
```
ahmedmuhsin, line 22: We now have two conventions for the same thing. `EventHubMessage.Body` uses `azfunc:"data"` and keeps a proper json name, while `ServiceBusMessage.Body` and `QueueMessage.Body` still use `json:"azfuncdata"`, so marshalling a ServiceBusMessage emits an `azfuncdata` key. Happy for it to be a follow up, but worth deciding now so it doesn't sit half migrated.

Unrelated to this line, one question on the PR overall. The runner still pins `TestPattern` to `^TestHttpTriggerGet$`, so none of this new integration coverage runs automatically. Deliberate for now?

sdk/bindings/cosmos.go
harshivcodes: Updated the PR description with a dedicated breaking-change and release-notes section. It now explains that Cosmos emits _ts as a JSON number, the previous string field broke real trigger decoding end to end, and consumers must migrate Timestamp usage to int64.

sdk/bindings/servicebus.go
harshivcodes: Updated the PR description to call out that binary dataType changes existing single-message behavior as well as batch behavior, preserving non-UTF-8 payload bytes. The empty-entry capability and alignment coverage remain documented and tested.

worker/converter.go
harshivcodes: Fixed in 884afe7. Metadata is now processed in two passes so per-message XArray values deterministically override scalar X values. Malformed or non-array extension metadata is skipped instead of failing and retrying the invocation, with regression tests for both cases.

sdk/bindings/eventhub.go
harshivcodes: Fixed the message-body convention in 884afe7: EventHubMessage, ServiceBusMessage, and QueueMessage now all use json:"body" azfunc:"data", with marshal regression tests. The integration runner pattern is deliberate for now: it only provisions Azurite for the HTTP milestone; the Event Hubs, Service Bus, and Cosmos tests require the explicitly started full emulator stack, which is now stated in the PR description and integration README.

gavin-aguiar (approved):

ahmedmuhsin (approved):
Approving. Re-verified 884afe7 against a real func host with the Service Bus, Event Hubs and Azurite emulators. All six integration tests still pass, and vet and unit tests are clean.

Both `collectionMetadataAt` issues are gone. I looped the scalar vs array precedence case 500 times and it is stable now, and malformed array metadata gets skipped instead of failing the invocation.

I also checked the body tag migration for fallout since it touches three message types. Single dispatch decode still works for `ServiceBusMessage` and `QueueMessage`, and a trigger metadata key literally named `body` does not clobber the raw input now that `Body` carries `json:"body"`. Both fine.

Thanks for pulling the breaking changes up into the description. The Cosmos one and the non UTF-8 one both needed to be visible to anyone upgrading.

